PR #3, opened 2016-02-08
Fixed: Changing data directory in GUI did not take effect if a record…

jonaskn:
…ing had been started (and stopped) before. Only the first directory before the first recording was used. Fixed: EventBroadcaster was linking to Debug version of open-ephys.lib in Release mode


jonaskn:
I moved the commits for the network messages into a separate branch. This should now be simple to pull

PR #5, opened 2016-03-05
FileReader: added possibility to drag&drop files. Refactoring and stylystic fixes.

sept-en:
Hello!

While working with app and testing I decided to add drag&drop support for file reader, so it will be much convenient to use for users.

Best regards,
Kirill Abramov.

### DIFF
--- a/Source/Processors/FileReader/FileReader.cpp
+++ b/Source/Processors/FileReader/FileReader.cpp
@@ -124,17 +124,35 @@ void FileReader::enabledState (bool t)
 }
 
 
+bool FileReader::isFileSupported (const String& fileName) const
+{
+    const File file (fileName);
+    String ext = file.getFileExtension().toLowerCase().substring (1);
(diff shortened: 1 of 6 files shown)


aacuevas:
Hi,

Could you please split this pull request in two, one with only the FileSource refactor and other with everything else?

Since FileSource is a base class for plugins, modifying the header declarations (even if it's just adding const to the const methods) breaks binary compatibility with already compiled plugins, which means we'd need to update the plugin api version macro and force rebuild them. Since the refactor doesn't changes actual functionality, I'd prefer to just save it for now and merge it later alongside any other future changes to the API.

Apart from that, Drag&Drop will definitelly come in handy, thanks!

Aarón.

sept-en:
Hello!

Okay, I've created new pull request without changes for FileSource class!
Check it, please!

Best regards,
Kirill Abramov.

PR #10, opened 2016-03-11
Development

priyanjitdey94:
Improvement of the channel selection interface. Text Box added in channel selection interface which now allows the user to select channels manually or by writing commands in the text box.


sept-en:
Check, please, if you've added your own changes in this pull request.
I can't see any your commits here.

PR #13, opened 2016-03-18
OS X packaging improvements

cstawarz:
As requested, this is a resubmission of open-ephys/GUI#324 for the new plugin architecture. The primary changes are as follows:

1. On OS X, Rhythm-related files (`libokFrontPanel.dylib`, `rhd2000*.bit`) are copied to and subsequently loaded from the application bundle's "Resources" folder.

2. Each currently-built plugin now has its own Xcode project, which builds the plugin into a loadable OS X bundle.  The projects get their settings from shared configuration files, so they can be re-configured in bulk by modifying those files.  To simplify the development process, there's also an Xcode workspace that includes the projects for the GUI and all plugins, with an "All" scheme that builds everything in one shot.

3. On OS X, plugins are installed in and loaded from the application bundle's "PlugIns" folder.

Also included in these changes are a few small fixes for compiler warnings from plugin code.

A few notes:

* The plugins that require third-party libraries (ZeroMQ, HDF5) explicitly link against the static versions, so the corresponding dylib's no longer need to be included with the OE app.

* In addition to looking for plugins inside the app bundle, PluginManager should probably also look for them in a per-user location (probably somewhere in `~/Library`).  This would make a better place for third-party plugins, as stuffing those in to the application bundle isn't ideal.  (For one thing, it would break [code signing](https://developer.apple.com/library/mac/documentation/Security/Conceptual/CodeSigningGuide/Introduction/Introduction.html) on distributed binaries.)

If/when these changes are merged, I will add a section to the [Plugin build files](https://open-ephys.atlassian.net/wiki/display/OEW/Plugin+build+files) article on the wiki that describes how to create new plugin Xcode projects.


jsiegle:
Thanks, this is a much-needed improvement! You can update the documentation, just make a note that the changes only apply to the "development" branch for now.

jsiegle:
I used the Xcode workspace to build all the plugins, and the .so files show up in the appropriate folders, but they're not being recognized by the app. Is there something I'm missing? Can you check if it's still working for you?

cstawarz:
Yes, it still works for me (except for the CAR plugin, which fails to load -- I'll look in to that shortly).  The plugins are no longer built as .so files, so presumably the ones you see are left over from a previous build.  All plugins (as well as required .bit and .dylib files) are now included in the open-ephys.app bundle.

My best guess is that you didn't select the "All" scheme before building, which is required to build the app and all plugins:
xcode_scheme

cstawarz:
> except for the CAR plugin, which fails to load

Fixed in #20

jsiegle:
Thanks, it works now! It was failing because I had the Open Ephys Xcode project open at the same time.

aacuevas:
About the per-user folder, would it use the same format as the bundled plugins or a different one? I'm not quite familiar with all this osx-specific stuff but, as you said, having a separate folder for third-party plugins (~/Library/open-ephys seems like a sensible option) is going to be needed sooner or later.

cstawarz:
> About the per-user folder, would it use the same format as the bundled plugins or a different one?

Yes, it would be exactly like `open-ephys.app/Contents/PlugIns`, except not inside the application bundle.  It seems like there's no real standard place to put this kind of thing, but I think `~/Library/Application Support/open-ephys/PlugIns` would be a good choice.

beOn:
Any update on the documentation for creating new plugin projects?

beOn:
There's a bit of legwork to be done if you're using homebrew instead of macports... changing /opt/local to /usr/local for a few of the bundles. Is hand-rolling projects for OS X plugin builds going to be the permanent approach?

cstawarz:
> Any update on the documentation for creating new plugin projects?

Sorry, I haven't gotten to it yet.  I'll try to put something up in the next few days.

> There's a bit of legwork to be done if you're using homebrew instead of macports... changing /opt/local to /usr/local for a few of the bundles. Is hand-rolling projects for OS X plugin builds going to be the permanent approach?

/usr/local does seem like a more sensible location, as it's the default for both Homebrew and a manual configure/make-install.  Unless there's some objection, I'd be happy to make that change.

As an alternative, you could make a symlink from /opt/local to /usr/local:

    $ sudo mkdir /opt
    $ sudo ln -s /usr/local /opt/local

beOn:
No rush - at least not from me. Got a plugin up and running, and the /opt/local stuff was easy to address.

cstawarz:
> I will add a section to the Plugin build files article on the wiki that describes how to create new plugin Xcode projects.

This is [done](https://open-ephys.atlassian.net/wiki/display/OEW/Plugin+build+files).  Sorry for the delay!

jvoigts:
This looks great, thanks!
i'll give this a try tonight, this is definitely going to be very helpful in making the whole build process easier.
We'll also start making a more visible list of what plugins are out there on the wiki and website soon.

jvoigts:
This looks great, thanks!
i'll give this a try tonight, this is definitely going to be very helpful in making the whole build process easier.
We'll also start making a more visible list of what plugins are out there on the wiki and website soon.

PR #19, opened 2016-03-22
Upgradation of Channel Selection Interface

priyanjitdey94:
Improvement of the channel selection interface:
1. Added a text box for selecting/deselecting range of channels.
Format used: [a:b]->select/deselect all channels from a to b
[a:k:b]->select/deselect channels from a to b such that each consecutive
selected/deselected channel has common difference k.
[a : : b]-> k=1 in this case.
[a: ]->select/deselect channels from a to last channel
[ : b]->select/deselect channels starting from 1 up to b.
[ : ]-> select/deselect all channels.
2.Update: Multiple ranges can be updated by using a comma separated list of ranges. 
For eg. [1:1:3],[4:7],[11: ]

### DIFF
--- a/Source/Processors/Editors/ChannelSelector.cpp
+++ b/Source/Processors/Editors/ChannelSelector.cpp
@@ -37,8 +37,8 @@ ChannelSelector::ChannelSelector(bool createButtons, Font& titleFont_) :
 {
 
     // initialize buttons
-    audioButton = new EditorButton("AUDIO", titleFont);
-    audioButton->addListener(this);
+	audioButton = new EditorButton("AUDIO", titleFont);
+	audioButton->addListener(this);
     addAndMakeVisible(audioButton);
     if (!createButtons)
         audioButton->setState(false);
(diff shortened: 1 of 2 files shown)


jsiegle:
Thanks, this is a really nice addition! @sept-en mentioned that the methods you added for parsing strings into ranges might be useful to other plugins as well. Can you try moving these into their own class?

priyanjitdey94:
Thanks! I was busy today. Sure I will move the methods into classes and submit it soon.

jvoigts:
While you're at it, I had the following suggestion on how to save some screen real estate without taking away any functionality:

By default, the channel selection should start somewhat scrolled down, so that only the lower half of the text input box is visible. As soon as the user clicks on the (lower half of the) text input box, or anywhere above the highest channel numbers, the scrolling is returned to 0, making the full text input box visible.

this way users who want to just be clicking on the numbers get to see an extra row of numbers, but users who want the text input need to click on it anyways so this doesn't add any extra clicks to their workflow.

priyanjitdey94:
@jvoigts Yeah sure. That will be an easy addition I believe. However there are many other ways if we consider an extra click.

priyanjitdey94:
@jsiegle Hi, here is the new class for parsing strings into ranges. [Link](https://github.com/priyanjitdey94/Online-Judge-Codes/blob/master/StringParser%20Class.cpp)

sept-en:
@priyanjitdey94 please, split the class into appropriate header and source files.

Also it makes a sense to:
- remove ctor/dctor as they are empty;
- it would be better to make methods as static since they don't require any class member data.
- "StringParser" class name is not very informative, I think something like "ListSliceParser" will be more convenient.

And I also think that it would be good if you can refactor ChannelSelector class according to the changes and then create pull request.


Best regards,
Kirill Abramov.

PR #21, opened 2016-04-02
New Class for converting string to range

priyanjitdey94:
@jsiegle @sept-en Here it is. Added a separate header and source file for converting strings to ranges.
Also, made changes to ChannelSelector.cpp accordingly as mentioned by @sept-en . Thanks.

### DIFF
--- a/Source/Processors/Editors/ChannelSelector.cpp
+++ b/Source/Processors/Editors/ChannelSelector.cpp
@@ -581,7 +581,7 @@ int ChannelSelector::getDesiredWidth()
 void ChannelSelector::buttonClicked(Button* button)
 {
     //checkChannelSelectors();
-    ListSliceParser* sp = new ListSliceParser("");
+    ListSliceParser sp;
     if (button == paramsButton)
     {
         // make sure param buttons are visible
@@ -686,7 +686,7 @@ void ChannelSelector::buttonClicked(Button* button)
         selectButtonParam->removeListener(this);
         deselectButtonParam->removeListener(this);
         int fa, lim, comd, i, j;
-        std::vector<int> getBoxList = sp->parseStringIntoRange(paramBox->getText().toStdString(),parameterButtons.size());
+        Array<int> getBoxList = sp.parseStringIntoRange(paramBox->getText(),parameterButtons.size());
         if (getBoxList.size() < 3)
         {
             selectButtonParam->addListener(this);
@@ -712,7 +712,7 @@ void ChannelSelector::buttonClicked(Button* button)
     {   // select channels in record tab
         selectButtonRecord->removeListener(this);
         deselectButtonRecord->removeListener(this);
-        std::vector<int> getBoxList = sp->parseStringIntoRange(recordBox->getText().toStdString(), recordButtons.size());
+        Array<int> getBoxList = sp.parseStringIntoRange(recordBox->getText(), recordButtons.size());
         int fa, lim, comd, i, j;
         if (getBoxList.size() < 3)
         {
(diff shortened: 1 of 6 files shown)


Source/Processors/Editors/ChannelSelector.cpp
```diff
@@ -580,6 +581,7 @@ int ChannelSelector::getDesiredWidth()
 void ChannelSelector::buttonClicked(Button* button)
 {
     //checkChannelSelectors();
+    ListSliceParser* sp = new ListSliceParser("");
```
sept-en: There we will get memory leak.

To get ride of this, the better way would be to define appropriate parsing method as static and just call it.
We don't need to create instance of this class.

Source/Processors/Editors/ListSliceParser.h
```diff
@@ -0,0 +1,51 @@
...
+3. [ x : p : y] -> stride parameter p. start from x, select channels at intervals p upto y.
+*/
+class ListSliceParser
+{
+public:
+    ListSliceParser(std::string s);
+    ~ListSliceParser();
+    std::vector<int> parseStringIntoRange(std::string textBoxInfo,int rangeValue);
+private:
+    std::string defaultString;
```
sept-en: Why do we need this variable? As I see we didn't use it at all, so it's better to remove it.

Source/Processors/Editors/ListSliceParser.cpp
```diff
@@ -0,0 +1,186 @@
...
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+*/
+
+#include "ListSliceParser.h"
+#include <vector>
+
+
+ListSliceParser::ListSliceParser(std::string s)
```
sept-en: I think we can remove constructor and destructor after we will make appropriate methods as static because it's just redundant.

Source/Processors/Editors/ListSliceParser.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+ListSliceParser::~ListSliceParser()
+{
+
+}
+
+int ListSliceParser::convertToInteger(std::string s)
+{
+    char ar[20];
+    int i, j, k = 0;
```
sept-en: As I said before, method convertToInteger() is deprecated, since it's better to use something like [std::stoi](http://www.cplusplus.com/reference/string/stoi/) because we have C++11 support.

Or do you have any reasons to use your custom method?

Source/Processors/Editors/ListSliceParser.h
```diff
@@ -0,0 +1,51 @@
...
+#include "../../../JuceLibraryCode/JuceHeader.h"
+
+class ListSliceParser;
+
+/*
+ListSliceParser Class: Find out ranges from a string separated by ';' or ','.
+The range can be of one of the following type:
+1. [ : ]        -> no start and end value. In this case start value is 1 and end value is index of last element
+2. [ x : y ]    -> range given. Start from index x and end at index y.
+3. [ x : p : y] -> stride parameter p. start from x, select channels at intervals p upto y.
```
sept-en: I would say it will be more convenient and more habitually if we use the simillar rules for slicing list to Python's.
In the Python for selecting some range of list we have the following structure (step parameter is the third).
[ beginIndex : endIndex : step ]

What do you think about it?

aacuevas:
Adding to all of @sept-en comments, just another minor one about our style standards.

Since all of our code is based around the Juce framework, for consistency and ease to combine with other parts of the code, as well as usually expanded functionality, we usually prefer the usage of Juce's own classes instead of the ones in the std library (juce::String instead of std::string, juce::Array<> instad of std::vector<>, etc...).

Of course, there might be cases when the use of the std methods are preferred, be it clear performance gains or better functionality but in general, Juce ones are preferred.

Source/Processors/Editors/ListSliceParser.h
priyanjitdey94: Hi, @jvoigts pointed out that the users of this applications are also matlab users. In matlab there is colon operator which works in a similar way.[See this](http://in.mathworks.com/help/matlab/ref/colon.html) 
So I think we don't need to change this.

priyanjitdey94:
@aacuevas  Thanks for pointing out. I was using std::vector and std::string for the same reason you mentioned and also because its easier to code. I changed it String and Array.
@sept-en Thanks. Made the changes. However, I don't want to remove convertToInteger method since it is also performing a check for invalid inputs( eg. large number that do not fit in 32 bit int  ) apart from converting string to integer. So, I kept it as it is. Keeping a separate module for converting strings to int will allow to implement additional checks as and when required.

PR #32, opened 2016-04-26
Mac path changes

cstawarz:
Changes to where various files are saved to, loaded from, and searched for on OS X:

* Look for user-installed plugins in `~/Library/Application Support/open-ephys/PlugIns`
* Default data directory is now `~/Documents/open-ephys`, which is created if it doesn't exist
* Store `lastConfig.xml` and `windowState.xml` in `~/Library/Application Support/open-ephys`
* Default directory for files explicitly saved/loaded by the user is now `~/Documents/open-ephys`

Also, the minimum supported OS X version is now set explicitly to 10.9.  (Without this, a build done on, say, 10.11 won't run on an earlier OS version).


aacuevas:
Thanks! User plugin paths for mac were much needed.

Could you please write some documentation on how to create project files for new plugins with this new build system, even if it's just a small tutorial? We're going to release a new binary build in a couple of weeks and we've asked other developers to start porting their old code to plugins, so we need them to be able to easily create new plugins.

cstawarz:
> Could you please write some documentation on how to create project files for new plugins with this new build system, even if it's just a small tutorial?

[Done](https://open-ephys.atlassian.net/wiki/display/OEW/Plugin+build+files)

aacuevas:
Cool! Thank you very much!

jvoigts:
Just added the lfp display node beta and it all worked flawlessly following your instructions - thanks!!

PR #43, opened 2016-05-20
Fix to get USB3 Opal Kelly device working on Linux

### LINKED ISSUE #44: Document how to get USB3 FPGA working on newer Linux distros
See #43 for a bugfix.

bburan:
* We need a newer version of the libokFrontPanel.so. This is copied from
  the latest version of the Front Panel SDK.
* The rhd2000_usb3.bit file needs to be in the same folder as the
  executable generated during the compile process.

Closes #44


aacuevas:
The correct FrontPanel library for USB support is already in /Resources/DLLs/Linux-USB3 as well as the bitfile in /Resources/Birfiles
The reason for them not being in the default folders by default is in #44

Issue #41, opened 2016-05-11
DataBuffer::addToBuffer numItems... useless, or just misunderstood?


beOn:
Unless I'm mistaken (definitely possible), the numItems argument from `DataBuffer::addToBuffer()` is confusing and potentially dangerous. Looking at the method's body (copied below) and calls to it throughout the code, the assumption seems to be that `numItems` will always be 1, as this is the case in all un-commented calls to it, and since `buffer.copyFrom()` gets called with `numSamples=1`.

The problem is, if you try to call `addToBuffer()` with anything other than `numItems=1` and one sample per channel in `data` argument, two unfortunate things happen (I think...):
1. `abstractFifo.prepareToWrite()` and `abstractFifo.finishedWrite()` get misinformed about the number of buffer positions being used (see (2)). Unless there's something going on I'm not appreciating, this will lead to gaps in the buffer that don't get written, but do get read.
2. The for loop (see below) only reads one sample per channel into the buffer, so if the method gets called with more than that, everything beyond the first sample is quietly ignored.

I'd be happy to contribute a version of `addToBuffer` that can take more than one sample per channel, and to provide a separate method that unambiguously does what the current implementation does (takes one sample per channel). It seems like it'd make `addToBuffer` safer and easier to use, and would be reverse compatible with any code that currently uses `addToBuffer` correctly (which includes everything in master). Whaddya think?

``` c++
void DataBuffer::addToBuffer(float* data, int64* timestamps, uint64* eventCodes, int numItems)
{
    // writes one sample for all channels
    int startIndex1, blockSize1, startIndex2, blockSize2;
    abstractFifo.prepareToWrite(numItems, startIndex1, blockSize1, startIndex2, blockSize2);

    for (int chan = 0; chan < numChans; chan++)
    {

        buffer.copyFrom(chan, // int destChannel
                        startIndex1, // int destStartSample
                        data + chan,  // const float* source
                        1); // int num samples
    }

    *(timestampBuffer + startIndex1) = *timestamps;
    *(eventCodeBuffer + startIndex1) = *eventCodes;

    abstractFifo.finishedWrite(numItems);
}
```

beOn:
Useless... no problematic... no, useless. So hard to decide!

aacuevas:
Behold what's probably one of the oldest classes in the entire GUI! Good catch on your part, it indeed has the issue you mention.

[...]

PR #47, opened 2016-05-25
addToBuffer with >1 item

beOn:
Addresses issue #41.


beOn:
Wait, there's an issue here. Fixing... will push soon.

beOn:
Okay. Much better.

PR #70, opened 2016-07-01
A better VisualizerEditor (deprecate buttonCallback)

arrow-:
* Added docs for the `VisualizerEditor` class. Override only `buttonEvent` _(as for other non-visual editors)_
* Started fixing other editors, a lot of redundant code can be removed.
* See comments in `VisualEditor.cpp`


jsiegle:
Thanks for this! Things are much simpler now.

PR #72, opened 2016-07-09
[VisualizerEditor] Helpers to override `buttonClicked` added.

arrow-:
These changes were made primarily to accommodate for the cyclops-stimulator. Since `buttonClicked` manipulates the `DataViewport` via `AccessClass`, a sub-class of `VisualizerEditor` cannot functionally override the method.
Hence:

* Added `protected` methods which manipulate `DataViewport` (via `AccessClass`)
* moved the SelectorButtons into `protected` area, essential!
* All of the additions _are_ essential to the cyclops-stimulator.
* There is NO change in the public interface.

Now, _my_ `buttonClicked` method can access the SelectorButtons, manipulate `DataViewport`.
Please remove the (new) comments from VisualizerEditor::buttonClicked (`.cpp`), they are not needed anymore.

### DIFF
--- a/Source/Processors/Editors/VisualizerEditor.cpp
+++ b/Source/Processors/Editors/VisualizerEditor.cpp
@@ -149,6 +149,11 @@ void VisualizerEditor::updateVisualizer()
 
 }
 
+void VisualizerEditor::windowClosed()
+{
+
+}
+
 void VisualizerEditor::editorWasClicked()
 {
 
@@ -191,8 +196,8 @@ void VisualizerEditor::buttonClicked(Button* button)
 
         if (dataWindow == nullptr) // have we created a window already?
         {
-            // dataWindow = new DataWindow(windowSelector, tabText);
-            makeNewWindow();
+            // enable windowClosed() callback
+            makeNewWindow(true);
             dataWindow->setContentNonOwned(canvas, false);
             dataWindow->setVisible(true);
             //canvas->refreshState();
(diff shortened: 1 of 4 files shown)


arrow-:
The cyclops-stimulator requires a shared canvas among it's editors, that is why I had to override `buttonClicked`

sept-en:
Hi, @arrow- .

Thanks so much for your contribution.
However, I recently did some work and refactoring for VisualizerEditor class in the branch for PluginGenerator (haven't pushed yet). So I'm aware about lot of conflicts, while merging later.

I guess we will merge in about a week **PluginGenerator** branch with **development** and I will definetely add your changes there, I think it's very useful.

Best regards,
Kirill Abramov.

arrow-:
@sept-en Ok!

Due to the last commit (windowClosed` callback), the `DataWindow` and `VisualizerEditor` have become highly coupled. That's in general a sign of bad design, but of the few ways I could do this, I think this is the _least bad_.

sept-en:
@arrow-, yeah I understand that some plugins might require this feature to somehow respond to the closing window events.

However, I think that's better approach would be to create DataWindow::LIstener and then subscribe VisualizerEditor for the events of DataWindow (Observer pattern). In my opinion, we will have a bit cleaner and less coupled design.

What do you think?

Best regards,
Kirill Abramov.

arrow-:
Thanks for pointing that out! I'll make that change, glad to not use `dynamic_cast`!

PR #73, opened 2016-07-26
Increased functionality of the Rhythm FPGA module

NewmanMemoryLab:
Our lab is using the DAC TTL functionality of the Rhythm FPGA to create a digital sync pulse between Open Ephys and other software (in our case, Bonsai). This update edits the display of the Rhythm FPGA module. 

What it does: Adds the ability to feed one of the input channels of the Intan board to one of the digital out pins on the board. A threshold in microvolts is set to determine what voltages of the input signal is determined "high" or "low", and the digital out pin will output the corresponding digital signal. (A positive threshold states that "high" is all voltages above that threshold, while a negative threshold states that "high" is all voltages below that threshold). Activating the DAC TTL button enables the DAC TTL functionality, while enabling a pin button activates the corresponding digital out pin. The channels are the input channels from the amplifier boards. While we're only trying to create a unique sync pulse, you can also use this to create a sync pulse that corresponds to neural spiking.

Due to space limitations, I've only added digital out pins 2-7 to be used, also because pins 0-1 can also be used for audio out. I'm fairly new to editing in C++, so any feedback is appreciated to make is more versatile or efficient.


aacuevas:
Hi,

Thanks for the code! We had at some point a similar functionality (thus the existing methods) but it wasn't complete, this is quite useful. I see, right away, three issues with the pull request, though. From less to more important:

-The pull request should be issued against the development branch, not the master one. All new development is being done in that branch while master is for tested versions of the GUI and the only commits that might go in there are critical fixes.

-As you yourself say, the Rhythm node editor ends being huge. Since the processor already has a canvas window for setting the name channels (can be opened by clicking the icons on the top right side of the editor) why not to move the DAC TTL selection interface to there? You'll have more space to work.

-And now the really important one: Do not call anything from `evalBoard `from a method that gets called when clicking a button. That includes the `setDACchannel` or `enableDAC` methods. The low-level funtions to handle the FPGA are not thread safe, as such, trying to change parameters while acquisition is running can (an does) cause issues, since the graphical interface and data acquisition tun on two different threads and the parameter command might be issued while the data transfer command is running, corrupting the FPGA state. As you saw, we have a `dacOutputsShouldChange` boolean variable to solve this: the GUI methods set some temporal variables and assert that and then all relevant parameters get updated in the `updateBuffers` method, just between calls to the data transfer command, to be sure that they do not collide. I don't know why you removed the code working that way from the already present methods but please rework them so they use that architecture, or any other that can prevent the race conditions.

Best,
Aarón.

PR #77, opened 2016-08-20
Sharp Wave Ripple Detector Plugin

annam95:
My lab is using open-ephys to investigate the role of [Sharp Wave Ripples (SWR)](http://onlinelibrary.wiley.com/doi/10.1002/hipo.22488/full) in a rodent neural circuit. We will use a similar closed loop system to the one described in [Siegle and Wilson](https://elifesciences.org/content/3/e03061). 

For this purpose, I developed a "SWRDetector" plugin. Much of the code for this plugin is based on the Phase Detector plugin. The algorithm used for SWR detection is based on [Stark et. al.](http://www.ncbi.nlm.nih.gov/pmc/articles/PMC4393648/):
- compute the RMS value of two running windows, long (2s) and short (8 ms), of band-pass filtered (80-250 Hz) signal
- when RMSShort > k \* RMSLong for continuous 8ms, trigger an event, with k a float constant input by user into GUI (threshold constant)
- event-on time can be input by user in GUI (event stimulation time)

I thought this plugin may be useful for other labs investigating SWRs, however, testing on its detection accuracy is so far limited. This is my first pull request, so any feedback on my code, or the way I submitted the pull request would be appreciated.


medengineer:
External plugins are now hosted on a separate repo.

zkcsmxz:
Could you tell me how to download and use this plugin, please? thanks a lot!

jsiegle:
Hi @zkcsmxz, this plugin never made it into the official released version of the GUI, but the source code is available here: https://github.com/mccannannamary/plugin-GUI/tree/master/Source/Plugins/SWRDetector

We've made a bunch of changes in the 4 years since the code was last updated, so it's not going to work as-is. We are now distributing selected plugins through a centralized repository, and this one is on the list of candidates to adapt and share. But we don't have an exact timeline for when this will be made available.

Issue #83, opened 2016-08-29
How do I build the GUI with Visual Studio 2015?


galenlynch:
I'm unable to build plugin-GUI with Visual Studio 2015 in Windows 10. I am unable to download visual studio 2013, and would like to test out some changes I've made before I submit a pull request.

I tried opening the 2013 solution and updating it, but kept getting MSBuild error MSB8020, where it could not find v140 build tools even though they were installed. I could compile a separate toy project, which suggests that the problem is with the environment variables or configuration in the project and not with my installation of Visual Studio.

Poking around at the visual studio configuration, it looks like it was generated not by visual studios but instead by introjucer/projucer? Is there some procedure for updating the visual studio project for visual studio 2015?

wehrlab:
I have similar issue with VS 2013 and other versions. If I clone the latest version of plugin-GUI, do I get updated code or do I need to do anything extra to add changes mentioned above?

galenlynch:
Hey sorry that I'm just now seeing this. My pull request was never merged, so you'll have to clone my fork if you want those changes. Note that you'll need to clone the `add-visualstudio2015-support` branch.

PR #85, opened 2016-08-30
Add Visual Studio 2015 support

galenlynch:
As it stands, I cannot compile the GUI using Visual Studio 2015 (#83). It looks like the Visual Studio solution and project file are generated by ProJucer? With these commits, I have updated the ProJucer file, and added the requisite resource to compile the main GUI program using Visual Studio 2015.

This is still not a complete fix to #83, because I don't know how to compile the plugins with Visual Studio 2015. The project files for each plugin also seem auto-generated by ProJucer, but I can't find the ProJucer file that generated them.


galenlynch:
Is this pull request not going to be merged? I'm happy to close it if you have some reason for not wanting to integrate the changes.

jsiegle:
@aacuevas can you take a look at? I don't think there's any reason not to merge it.

aacuevas:
Oh, I missed this one. No, there is no big reason not to merge this. However, Galen, could you please create the appropriate VS2015 project files for the plugins? I don't have 2015 installed right now and, since the plugins need to be built with the same compiler as the base GUI, ideally we should ship projects for them as well.

galenlynch:
Sure thing! I'll only be able to do it early next week, however.

PR #94, opened 2016-10-13
Mac build updates

cstawarz:
These changes fix some macOS build issues and update some Xcode project settings.  They also add Xcode projects for the JuliaProcessor and BinaryWriter plugins.


aacuevas:
Chris, is there any specific reason XCode is so picky about the order of member initialization in the constructors? What is the rule, so we can try to write them in the right order in the future? It doesn't seem to be an issue on linux or windows.

cstawarz:
Xcode (well, really clang) is complaining because the project files include `-Wreorder` in their compiler options.  This flag is supported by gcc, too.  [This Stack Overflow question](http://stackoverflow.com/questions/1828037/whats-the-point-of-g-wreorder) has a succinct explanation of what it does and why it may matter.  I don't know why only the Mac build uses this flag.  (I didn't add it, though I think it's a good idea.)

To avoid triggering the warning, the order of the member initializers must match the actual initialization order.  For non-static data members, this means the order in which the members are declared in the class definition.  See the section "Initialization order" in  [Constructors and member initializer lists](http://en.cppreference.com/w/cpp/language/initializer_list) for more details.

aacuevas:
Oh, so the order in which members are initialized is not the order they appear in the constructor but the one they are declared in, interesting, I didn't notice that particular detail. (not quite intuitive, but in line from what one might expect from C++) Makes sense if you think that that way all members are initialized in the same order regardless of what different constructors might do.

Thanks for the info!

Issue #89, opened 2016-09-05
plugins, translation units, and multiple instances of LeakedObjectDetector's LeakCounter singletons


beOn:
I thought this was going to be a simple bug...

Here's the issue: 
- Build the latest development branch (I'm on a mac using xcode, but this is probably a cross-platform issue).
- Run it.
- Throw a Spike Viewer in the signal chain.
- Open the Spike Viewer's tab, make sure it has focus, then press a key on your keyboard.
- **You'll hit a breakpoint**, and find this waiting for you in the debug console:

```
*** Dangling pointer deletion! Class: KeyPress
JUCE Assertion failure in juce_LeakedObjectDetector.h:71
```

Bummer! But Ok, I thought, let's fix it. And that's when my weekend plans went out the window.

Long story short, usually the LeakedObjectDetector class template uses one LeakCounter singleton per class that uses the template. So in this case, there's supposed to be one LeakCounter dedicated to counting constructions and deletions of KeyPress objects. But it turns out that singleton's don't behave as you might expect when combined with a plugin architecture - each plugin (and the main app) has its own compilation unit (translation unit), and, as things stand now, we actually get one LeakCounter singleton per class that uses the LeakedObjectDetector template _per translation unit_. That's crazy. It makes sense, but ugh, crazy. Let's take a closer look at the craziness, just to be sure.

First things first. The issue goes away if you comment out the entire body of `SpikeDisplayCanvas::keyPressed`. If you then un-comment only the line

```
KeyPress c = KeyPress::createFromDescription("c");
```

... the issue returns. So, simply instantiating then deallocating a KeyPress within this method causes a false dangling pointer alert. That's weird. But check out what happens if I modify getCounter to talk to me a bit (code borrowed and modified from [here](https://forum.juce.com/t/solved-has-anyone-been-able-to-build-the-rtas-sdk-with---cdecl-yes-the-doc-says-it-is-possible/17919)):

``` diff
--- a/JuceLibraryCode/modules/juce_core/memory/juce_LeakedObjectDetector.h
+++ b/JuceLibraryCode/modules/juce_core/memory/juce_LeakedObjectDetector.h
@@ -107,6 +107,13 @@ private:
     static LeakCounter& getCounter() noexcept
     {
         static LeakCounter counter;
+ 
+        static int LeakCounterCounter; 
+ 
+        if (strcmp (getLeakedObjectClassName(), "KeyPress") == 0) { 
[...]

PR #95, opened 2016-10-14
Symbol visibility fixes

cstawarz:
These changes resolve issue #89 on macOS by keeping inline functions visible in debug builds.  They also ensure that plugin entry-point functions are always visible on macOS and Linux.


aacuevas:
Oh, I don't know how I didn't notice the export visibility attribute before (when I did very same thing for the PLUGIN_API macro) Thanks!

beOn:
Yup, works for me to. Thank you!

Issue #42, opened 2016-05-19
GUI crashes when dialog in File Reader plugin is opened on some Linux systems


bburan:
To quote a post on the forum:

> the issue is probably because of a weird zenity (the native dialog box utility for many Linux variants) issue. We can build the GUI to not use zenity for choosing files, as follows:
> 
> Navigate to the Builds/Linux folder (you might want to take a backup of the Makefiles, but don’t backup the Builds/Linux/builds/*).
> 
> Open the Makefile and look for the lines (there should be 2) that say
> `CPPFLAGS := $(DEPFLAGS) -D “LINUX=1” -D “DEBUG=1” .....`
> 
> Now add anywhere the macro definition `JUCE_DISABLE_NATIVE_FILECHOOSERS=1`(the line may not be exactly same, you’ve to add the new macro definition to the `CPPFLAGS` variable) `CPPFLAGS := $(DEPFLAGS) -D “LINUX=1” -D “JUCE_DISABLE_NATIVE_FILECHOOSERS=1” -D “DEBUG=1”`
> 
>    Do the same for Makefile.plugins.
> 
>   It’s strongly recommended that you use the Introjucer for plugin development, but it suffers from the same problem (build will pass, GUI will launch, but File Chooser will hang the GUI). You’ll have to edit the makefiles in Introjucer/Builds/Linux in exactly the same way.
> 
>   To make this kind-of persistent, whenever you open OE with Introjucer, just make sure the `"JUCE_DISABLE_NATIVE_FILECHOOSERS=1"` is present in the “Extra Preprocessor Definitions” of the Config tab > Linux Makefile, as shown. Now all plugins and the GUI would get this macro definition by default.

See https://groups.google.com/forum/#!topic/open-ephys/NydCGT48tF0 for the source of this quote.

malfatti:
This fixed all problems I had in gentoo (https://github.com/open-ephys/plugin-GUI/issues/34). Thanks!

arrow-:
As I've mentioned in the forum, the bug seems to have roots in the way JUCE invokes zenity. See this [thread](https://forum.juce.com/t/freeze-when-opening-filechooser/16026/15) on the JUCE forum. They've very recently managed to squash this bug and the patch is on their develop branch. I believe by the time OE switches to JUCE 4.X, it would be pulled into master _(I'll keep track of this too)_, and we can close this issue.
Till then, we have to use the non-native dialog box.

malfatti:
[...]

PR #100, opened 2016-11-04
In order to correct dialogbox problem on xubuntu 16.04 (filereader, save config, etc.)

delcasso:


arrow-:
This is a quick fix for #42. I still haven't got around testing the latest JUCE 4.2.X _(nor 4.3.0)_ which, I believe have resolved the issue (see comments on #42).
A simpler solution is applied by @ablot [here](https://github.com/ablot/plugin-GUI/commit/e4bcf65dbb445660c4e5f665ff0857f3ccacb90e), by just adding a compile flag that disables native dialog boxes completely.

aacuevas:
Is this issue still present in the development branch of the GUI? We updated its Juce version a bit ago but it still hasn't made the jump to the master branch.

delcasso:
Yes !

arrow-:
Ok, I dug into JUCE history.
@aacuevas The upstream patch was added on May 21, 2016 (on JUCE develop branch). So I expect JUCE 4.2.3+ to have the patch.
And (I think) we are on JUCE 4.2.1, see 3f5f403bfe484d1a5e043bb320ca24e6c938e186

delcasso:
@aacuevas Sorry for the brief answer. I tested the development branch and that was still the case.
I you need I could make more tests.

aacuevas:
It shouldn't be too hard to update JUCE to 4.2.3. In the meanwhile, this patch is still useful.
I just have two concerns I'd need to ask before merging:
-Since the proposed patch just disables the native file chooser for all linux builds, but the bug seems specific to xubuntu, and unaffected users might prefer their native dialogs, perhaps it would be better to follow @ablot's solution linked by @arrow-, since compiler flags are easy to modify.
-Usually patches go to the development branch, but this seems an important enough fix to go directly to the master branch. Our policies on this kind of things are kinda new so I'll have to check.

It seems I cannot reopen the pull request, but once those topics are talked about I'll implement one of the versions of this fix, until we can update Juce.

delcasso:
Hi Aaron,

I modified the compiler flags in the makefile, it's cleaner. We should 
probably update the linux-installation-procedure with a link to another 
Makefile (for xubuntu and gentoo).

We could also generate binaries for xubuntu.

Sebastien

I you need I can remoake my changes on my for and creates a pull request

PR #101, opened 2016-11-12
Update ExampleProcessor.h

delcasso:
To fix error during running open-ephys

Loading Plugin: ExampleProcessor ... ../../Source/Processors/PluginManager/PluginManager.cpp:177: Failed to load plugin DLL: /home/username/plugin-GUI/Builds/Linux/build/plugins/ExampleProcessor.so: undefined symbol: _ZTV19ExampleProcessorEditor
 DLL Load FAILED

### DIFF
--- a/Source/Plugins/ExampleProcessor/ExampleProcessor.h
+++ b/Source/Plugins/ExampleProcessor/ExampleProcessor.h
@@ -50,7 +50,7 @@ public:
     ExampleProcessor();
 
     /** The class destructor, used to deallocate memory */
-    ~ExampleProcessor();
+    ~ExampleProcessor(){};
 
     /** Determines whether the processor is treated as a source. */
     bool isSource()


aacuevas:
If the error is with ExampleEditor wouldn't this patch need to be in ExampleEditor.h instead of ExampleProcessor.h? The ExampleProcessor destructor is already defined in ExampleProcessor.cpp, but the ExampleEditor isn't, which is the cause of the issue you describe.

In any case, it would be better to define the destructor (even an empty one) in the .cpp, same as we do with the processor. It's not too important, but helps avoiding possible rare issues.

delcasso:
Oh Yes !!

Sorry, the patch need to be in the ExampleEditor.

On xubuntu when I created a new plugin with an editor, i didn't show up 
in the gui because of this error.

Thanks Aaron.

Sebastien

PR #117, opened 2017-02-15
Fix eventData size limit of 255 bytes

mspacek:
This fixes an overflow that could be easily triggered by sending a long message via the message center (bottom bar in the GUI), resulting in only some fraction of the message being written to the messages file. Note that this triggers a recompile of a bunch of (presumably dependent) modules. Not sure if changing the data type of numBytes from uint8 to int has some other knock-on effects that I'm unaware of. Seems to work, so far.


aacuevas:
This is the kind of changes I need to triple-check to be sure it doesn't break something obscure, but thanks for the patch.

However, I won't be too worried about events right now. I've been working on a revamp of the channel and event systems which is almost ready to be merged into development.

PR #119, opened 2017-02-20
Fix backwards RISING and FALLING Edges enum in RecordControl

mspacek:
I copied the following line from `RecordControl.h` into my own plugin to detect rising and falling edges:
```
enum Edges { RISING = 0,  FALLING = 1 };
```
I thought the choice of values was a bit weird, but arbitrary, and as long as it works, it doesn't matter. However, after a bit of testing with TTL events (and saving entire TTL words to disk on bit change), I eventually found that the eventID of rising edge TTL events is actually 1, and of falling edge events is 0. So, the enum in RecordControl has it backwards. I couldn't find any other place in the code that treated the TTL eventIDs backwards this way, so hopefully it's the only place that got it wrong.

### DIFF
--- a/Source/Plugins/RecordControl/RecordControl.h
+++ b/Source/Plugins/RecordControl/RecordControl.h
@@ -53,7 +53,7 @@ public:
 private:
     int triggerChannel;
 
-    enum Edges { FALLING = 0, RISING = 1 };
+    enum Edges { RISING = 0, FALLING = 1 };
     enum Types { SET = 0, TOGGLE = 1};
 
     Edges triggerEdge;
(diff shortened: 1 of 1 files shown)


mspacek:
Note that this bug also exists in master.

aacuevas:
This is not much of a bug but perhaps more of a readability thing. We never check the eventID against that particular enum, there is a line with `int edge = triggerEdge == RISING ? 1 : 0;` to get the actual value that is checked against, so no matter the values of the enum, which was never intended to be used outside the RecordControl processor, the check is being done correctly.

In fact, inverting just this particular enum will cause the processor to run backwards, since the ComboBox selector still sends a "0" for rising and a "1" for falling edge to the processor (this is related to the way comboboxes work and how we pass parameters from the editor to the processor, having "rising" as the first option in the selectos does usually have more sense).

mspacek:
Ah, I guess I went looking through the code to figure out what eventID meant for a TTL event, found this, and misinterpreted it. Oops!

PR #120, opened 2017-02-20
Update setup.sh to use new Makefile

FlorianFranzen:
These two simple patches are for the linux setup.sh script:
- Link the correct binary to ```/usr/bin``` 
- Use new ```Makefile.plugins``` to build plugins as described in the wiki


aacuevas:
Well, the .so thing in the open-ephys binary file is a bug somewhere in the jucer options, it should be "open-ephys" withouyt any extension. I'll check that out before merging this.

PR #123, opened 2017-03-07
add destructor to ExampleEditor.cpp

e-matteson:
This is a followup to PR #101, because it doesn't look like that one got updated or merged. I had the same problem with being unable to load an example plugin, and I think this fixes it. Let me know if the destructor should actually do something in this example instead of being blank.


medengineer:
Closing because this file has new location and destructor already exists.

PR #145, opened 2017-07-21
Voltage crossing detector plugin

ethanbb:
Hello maintainers,

I'd like to contribute a plugin that I've developed for use in the TNEL at MGH, but should be generally useful for various purposes. The purpose is to fire a TTL event when a specified input data channel crosses a specified threshold level, with either a positive or negative slope (or either); the critera for detection and the output are highly customizable. It does not modify the continuous channels. Here's what it looks like, and a shot of it in action processing data with the settings shown:

crossingdetector2
crossingshot

How it works:
- With the default settings, a positive ("rising") crossing occurs at sample _t_ if and only if sample _t_-1 is less than the threshold and sample _t_ is greater, and vice versa for a negative ("falling") crossing. However, to make it more robust to noise or just tweak it to fit a certain use case, you can also adjust the span and strictness settings:

  - **Span** controls how many samples before or after the current sample are considered.
  - Of these, the percent that must be on the right side of the threshold is controlled by **strictness**.
- The duration of events, in samples, can be adjusted ("dur").
- **Timeout** controls the minimum number of samples between two consecutive events (i.e. for this number of samples after an event fires, no more crossings can be detected).

I've tested it on Windows and Linux, but I don't have access to a Mac so there is currently no OSX project. It shouldn't be too hard to set up though, since it's a standard plugin that requires no external libraries.

Let me know what you think!


ethanbb:
Closing to post on the [3rd-party plugin repositories](https://open-ephys.atlassian.net/wiki/display/OEW/Third-party+plugin+repositories) page instead.

Issue #90, opened 2016-09-06
How to link multiple plugins with common libraries?


ethanbb:
This issue has been brought up in PR #57, and I'm giving it a separate thread here.

In the process of developing a new plugin, I've realized that I need to give it some of the filtering functionality that is currently available to the FilterNode plugin through the [Dsp library](https://github.com/vinniefalco/DSPFilters), which lives in the FilterNode source directory. However, I don't want to duplicate code, and I'm fairly sure that copying-and-pasting this library into the source of another plugin would create namespace clashes and linking errors. Thus, I'm trying to figure out how one could share a library between multiple plugins without disrupting the plugin architecture. 

It seems that the natural solution would be to compile this library into a separate shared library file (.lib/.so/.dylib), and then have it be available when compiling the plugins that need it (and do the same with any other libraries that need to be shared). Since it would be messy to try to make that work with the way plugins are compiled now, maybe a better way is to somehow incorporate needed libraries into the main OpenEphys build? I know NetworkEvents, EventBroadcaster and I think one other plugin share ZeroMQ and get around this by requiring ZeroMQ to be installed as a prerequisite (at least on Linux?) but that obviously isn't an option for all libraries.

Thanks,
Ethan

aacuevas:
Having a plugin require a specific library code in the GUI defeats the purpose of the plugins, which is them being independent.

While this might be useful for specific code that can benefit many plugins (The Dsp library being an example. In fact I tried to have it live inside the GUI when we switched to plugins but I ran into issues doing so) it is not the solution for common code specific to a couple of individual plugins.

The ideal solution is making the common code into a shared library. It is not really messy to use them with the current plugin build system, is just a matter to add yet another dependency, as the plugins using ZMQ or HDF5 libs do, but we should find a way to streamline the creation of the build files, as that's the really dull work. @sept-en, how difficult do you think adding another build option to the plugin generator to create a simple empty library template and build project files would be?

ethanbb:
OK, I agree that adding code to the main GUI library for the benefit of specific plugins doesn't make much sense.

I didn't realize that the plugins using ZMQ and HDF5 were using shared libraries - where is that in the source code? All I can find is that in the Windows build folder, there's a ZMQ dll in the four main bin directories.

ethanbb:
[...]

PR #146, opened 2017-07-21
Make DSP a common library for plugins

ethanbb:
This change does not modify the behavior of the program in any way; it is just an architectural change that addresses the issue I was having in #90 by making the digital signal processing library used by FilterNode into a shared library that can be reused by other plugins. This helps me with developing another plugin that's using this library, and I'm sure others would find use for the various types of filters it provides as well. (Thanks to @aacuevas for setting up the common library template, by the way!)

It definitely compiles on Windows and Linux, but I have not touched the OSX build since I don't have access to a Mac.


ethanbb:
I'd like to get this merged if possible - it shouldn't change any behavior, and I need the DSP library to be shared for a plugin I'm planning to release soon to work. Plus, then it's available for other plugins to use. I know it would break the OSX build though - is anyone willing to try to port the changes? Otherwise I can try to get access to xcode.

aacuevas:
Done, I have managed to build and test the shared lib in mac. It seems to be working with no issues.
Thanks for this and sorry it took so long!

ethanbb:
Thank you!

PR #167, opened 2017-10-17
EvntTrigAvg

Claybarn:
Addition of the EvntTrigAvg module. Aligns spikes from each electrode to events by default, and by sorted Id when they are added. The window size and bin size are adjustable.


jsiegle:
Can you merge this into open-ephys:development instead?

aacuevas:
I'm closing this, since the plugin is alreadyintegrted in the testing/devel branch .

PR #170, opened 2017-11-05
Edit all thresholds button

Claybarn:
Added button to allow user to edit all thresholds simultaneously. Also fixed some crashing issues when dragging the threshold bar. However, crash will still occur when changing electrodes.


aacuevas:
This seems to have been based off an older version of the code, which makes it impossible to automatically merge due to various conflicts. Could you please check it out and merge the code on your end with the latest code base?

Claybarn:
Done! Looks like I had messed with some things and forgot to revert. All good now!

PR #179, opened 2017-11-07
Add X11 window size hint to juce library code for minimum window dimensions

kmichaelfox:
These dimensions match the minimum size of the Windows platform.


aacuevas:
I don't suppose there is any way to solve this without modifying Juce code, right? Can you please mark the code within comments with the word open ephys on them, so in case we update juce can keep a good track of any changes we made to it?

kmichaelfox:
The latest commit has the section commented

PR #180, opened 2017-11-07
Adding VisualStudio2015 build files

kmichaelfox:


aacuevas:
Are the project files for the main GUI generated manually or from Projucer? I should be the latter, as it allow for easy configuration of all platforms.

Also, have you tested HDF5 plugins (like KWIK or NWB format or the KWD file reader) with this? The main thing holding back an update of the VS tools was compatibility with the hdf5 lib.

kmichaelfox:
These VS2015 build files were just duplicated from 2013 and updated for VS2015. I wasn't aware that 2015 tools were waiting on HDF5, so those aren't tested in here.

aacuevas:
Yeah, basically: The DLLs for the version of HDF5 we're using aren't compiled for VS2015, the newer ones aren't compatible with our current code but they are not available on the repositories of all linux distros.
So unless the VS2013 hdf5 dlls worked with a vs2015 build, which if you could test would be great, the plan is to build the dlls for our current version ourselves, before making VS2015 100% official.

kmichaelfox:
I think this should be closed and not pulled. There is too much risk of untraceable bugs emerging from toolchains.

PR #184, opened 2017-11-14
Fix bug with DataChannels and audio monitoring

kmichaelfox:
DataChannel objects are currently copied when they are prepped for acquisition. Also, the copy constructor for DataChannel did not (before these commits) inherit the recording, enabled, or monitored state from the original when the original is copied.

Changing the state _during_ acquisition resulted in changing the copy as desired, but the original DataChannel's state was left unchanged. Stopping acquisition and resuming would not be up-to-date with user configuration during the last acquisition. Changing state before enabling acquisition would not propagate to the copy, because the copy constructor hardcoded the value. Both of these issues were changed in this commit.

Since RecordNodes are handled in the same way, it is likely that this bug exists there too. However that behavior has not been changed yet.

Other stray changes:

Minor updates to MacOS .gitignore rules. A few new things to ignore, and some more generalized versions of existing rules. The generalized ignore rules have, for now, been included along with their more specific rules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,13 +120,35 @@ Builds/MacOSX/open-ephys.xcodeproj/*.pbxuser
 Builds/MacOSX/open-ephys.xcodeproj/xcuserdata
 Builds/MacOSX/open-ephys.xcodeproj/project.xcworkspace
 Builds/MacOSX/build
-Projucer/Builds/MacOSX/The Projucer.xcodeproj/*.mode1v3
-Projucer/Builds/MacOSX/The Projucer.xcodeproj/*.pbxuser
-Projucer/Builds/MacOSX/The Projucer.xcodeproj/xcuserdata
-Projucer/Builds/MacOSX/The Projucer.xcodeproj/project.xcworkspace
+Projucer/Builds/MacOSX/Projucer.xcodeproj/*.mode1v3
+Projucer/Builds/MacOSX/Projucer.xcodeproj/*.pbxuser
+Projucer/Builds/MacOSX/Projucer.xcodeproj/xcuserdata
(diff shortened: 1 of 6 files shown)


aacuevas:
Not copying the recording or monitoring status is by design. Channel copying is primarily done when updating the signal chain, with nodes down the chain copying the channels of their ancestors. If those status were copied, all processors would inherit the status of the previous processors, which can cause all the processors in the chain to record, while only the first one is intended to.

Record and audio nodes work a but differently, creating their local copies when acquisition starts. The record node does update the status of its local copies, reflecting the value of the original ones. The audio node does not, so I'd suggest changing that in AudioNode::addInputChannel instead of the standard copy constructor.

Rearding changing the recording or monitoring status while acquiring, you're right, the status wouldn't hold between acquisitions, so the local channel flag should be also updated.
Note that recording status changes shouldn't be permitted during recording (but should be during acqusition, between recordings) and RecordNode::setChannelStatus fails when trying to do so. Local channe recording flag shouldn't change if that were the case. Perhaps the best option for this is making RecordNode::setChannelStatus to return a boolean and check the return value before setting the local recording flag.

kmichaelfox:
Thanks for the explanation. I'll make the updates accordingly.

kmichaelfox:
The latest commit removes the state inheritance from the copy constructor as outlined above. AudioNode copies are manually set to the value at the time of copying in AudioNode::addInputChannel.

Revisions to RecordNode required a few updates. Most of them are what you described before. The ChannelSelector will also now check the resulting state of a DataChannel after attempting to override it, and manually forcing the button to match that state. This will prevent the button state and associated RecordNode state from being out of sync, such as when a state change during active recording is blocked.

PR #189, opened 2017-12-15
Binary Writer (10-Bit)

kmichaelfox:
This is a plugin that provides an alternative binary writing format that packs 10-bit sample values into 16-bit integer arrays without padding. A total of eight 10-bit samples are stored within five contiguous 16-bit integer values.
       
E.G.  | XXXX XXXX : XX00 0000 | 0000 BBBB : BBBB BBEE | EEEE EEEE : ...
(where | demarcates integer addresses, : demarcates byte addresses within integers)

The 10-bit binary writer retains much of the structure of the original binary writer. Floats are converted to 16-bit LE values and packed into per-channel collections then passed to SequentialBlockFile. The significant difference is that the conversion from float to int16 bitmasks the 10 most significant bits and performs the 10-bit compressed packing. After the packing, the per-channel data is interleaved per-byte and not by per-sample.


jsiegle:
I think we want to keep this as a separate repository for now, since it's only useful for the Neuropixels recordings. Can you create a new repo that I can fork?

PR #192, opened 2017-12-27
Some more Pulse Pal fixes

ethanbb:
Hi,

Changes to the Pulse Pal Output project:

* Added canvas sources to the "source files" for VS so it would compile
* Added loop variable initializations in PulsePalOutputCanvas.cpp


ethanbb:
Got rid of the visual studio file changes since they've since been fixed on development. It looks like these changes in PulsePalOutputCanvas.cpp are still necessary though - I get a runtime error without them when opening the Pulse Pal window or tab in Windows.

ethanbb:
That was fast!

PR #194, opened 2017-12-29
Merge updates on Master back into Development

kmichaelfox:


aacuevas:
I already did this manually, I think almost at the same time you issued this. Right now the branches are synced, so I'll close the request. Thanks anyway!

PR #207, opened 2018-04-13
Fix invisible channel rows issue in Channel Mapper

ethanbb:
This change fixes an issue where any non-full rows of channels would not be visible in the Channel Mapper (e.g. if there are 35 inputs, the first 32 channels show in 2 rows of 16, but the last 3 channels are invisible). As far as I can tell, the issue was limited to these channel buttons not being visible; with this fix they should have all the functionality of the other, previously visible channel buttons.


mspacek:
Nice catch. I assumed the aux and adc chans were left out of the chanmap on purpose. I guess not!

This patch seems to work fine for me. Here's what it looks like when applied against #227 with an A1x64 channel map loaded, and aux and adc chans toggled a couple of times:
chanmap_rows

PR #211, opened 2018-04-20
Sample math plugin

ethanbb:
The sample math plugin performs simple arithmetic on its inputs, adding, subtracting, multiplying or dividing by a constant or one of the inputs. I already posted it on the 3rd-party plugins page, but I'm putting this here just in case you think it's worth including as a default plugin. Some example uses:

* Subtracting out a reference (overlapping with Common Avg Ref functionality)
* Demeaning and normalization (using running mean and std dev plugin)
* Increasing dynamic range of low-intensity signals, for more precise recordings
* Scaling a control input to another plugin (I don't think this applies to any built-in plugins, but that's what we're planning to use it for)

sm_chan sm_const

I totally get it if you don't want to add any more defaults - just thought that if you do, this might be a good candidate since it's pretty general-purpose.


ethanbb:
Closing because I'm no longer keeping this branch up to date.

PR #212, opened 2018-04-30
Fix 'drawableSubprocessor' not being set when not loading settings or explicitly selecting a subprocessor

ethanbb:
Currently, the LFP viewer has a bug where the canvas's `drawableSubprocessor` field doesn't get set in the typical case of adding a new LFP viewer to a signal chain and not selecting a subprocessor from the ComboBox. The `updateSubprocessorSelectorOptions` method selects the default subprocessor to display in the ComboBox, but doesn't send a notification, which is a problem because the ComboBox listener is responsible for setting `drawableSubprocessor`. The end result is that the canvas does not refresh properly and the data does not get displayed.

This fixes the bug by making `updateSubprocessorSelectorOptions` send a notification, and getting rid of a resulting circular call to `updateSettings` which should never be necessary, given the contexts in which the containing function `LfpDisplayNode::setSubprocessor` is currently called. The other change just removes a call to `updateSubprocessorSelectorOptions` which seems to be redundant with line 88 in that file.

### DIFF
--- a/Source/Plugins/LfpDisplayNode/LfpDisplayEditor.cpp
+++ b/Source/Plugins/LfpDisplayNode/LfpDisplayEditor.cpp
@@ -71,7 +71,6 @@ void LfpDisplayEditor::stopAcquisition()
 Visualizer* LfpDisplayEditor::createNewCanvas()
 {
     canvas = new LfpDisplayCanvas(lfpProcessor);
-    updateSubprocessorSelectorOptions();
     return canvas;
 }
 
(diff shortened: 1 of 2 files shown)


ethanbb:
No problem! Now just trying to figure out why TTL events sometimes aren't showing up...

jsiegle:
What does your subprocessor configuration look like? @aacuevas created a source with reconfigurable subprocessor streams/sample rates that could be useful for testing.

ethanbb:
Actually never mind, looks like my problem was that I had the event channel turned off in the LFP viewer settings...thanks for helping me take a step back, I was getting pretty frustrated!

PR #220, opened 2018-05-15
Fix bug when loading LFP viewer with no channels

ethanbb:
Currently, the GUI crashes when trying to load a config file that contains an LFP viewer with no input channels. This fixes the bug, which seems to have been caused by missing curly braces on a multi-line if statement.


ethanbb:
Hey, bumping this - it's a super simple fix for something that's clearly a bug and has high impact.

Issue #199, opened 2018-02-21
Selected reference channel is not saved


jonnew:
OS: Ubuntu 16.04
Version: 8406ede366ef025c8754f17e56eb52510739524e

When I save a OE GUI settings file, and reload it, the previously selected reference channel (in 
"param" section of the Channel Map component) is no longer selected or applied.

ethanbb:
Closed by #221 (I don't have the permission to close this but it should be fixed)

jsiegle:
The Channel Map plugin no longer includes a digital referencing option. Instead, we recommend using the Common Average Reference plugin, which should save/load all settings exactly as they were when the GUI was closed.

PR #221, opened 2018-05-16
Save gain, reference and affected channel state in CAR plugin

ethanbb:
What it says on the tin - save the Common Avg Ref parameters to XML (gain in an editor custom attribute and reference/affected channels in each channel's custom attributes).

### DIFF
--- a/Source/Plugins/CAR/CAR.cpp
+++ b/Source/Plugins/CAR/CAR.cpp
@@ -138,3 +138,43 @@ void CAR::setAffectedChannelState (int channel, bool newState)
         m_affectedChannels.add (channel);
 }
 
+void CAR::saveCustomChannelParametersToXml(XmlElement* channelElement,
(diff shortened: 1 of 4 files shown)


jonnew:
This has been true for ages. If ppl do their recordings following a channel map (and therefore referencing), this can totally screw up their data in an irreversible way.

ethanbb:
Yeah, sorry. I misread that other issue, didn't realize it was about something else which still isn't fixed.

jonnew:
Actually, I might be confused. I didn't realize I was looking at PR I thought I was in issues still. Let me try to compile development and see if its actually fixed.

jonnew:
OK, yeah so this is specifically for the CAR plugin, correct? The channel map also does not save reference channels which was what #199 refers to.

ethanbb:
Yup, exactly.

PR #225, opened 2018-06-06
 Label LFP Viewer chans by name instead of number

### LINKED ISSUE #224: Channel labels don't propagate between blocks
I only recently realized this after messing around with the channel map plugin and noticing it wasn't changing the labels in the LFP viewer. From the channel block's [help page](https://open-ephys.atlassian.net/wiki/spaces/OEW/pages/950421/Channel+Map)

> To rearrange channels, click REMAP (it will become darker). You can then re-order channels by dragging. In subsequent blocks, "channel 1" will refer to the first channel in the list (in the example picture above, channel 2).

IMO, this is a fatal flaw, and makes open-ephys nearly unusable. How is anyone supposed to know what they're looking at in e.g. the LFP Viewer if the "channel numbers" are always labelled 1..nchans, regardless of how they were rearrange by the preceding channel map block? After staring at the code, I'm not sure the infrastructure is even there to propagate actual channel numbers (i.e. labels) between blocks, instead of always recalculating channel indices for each block. Is this feasible?

mspacek:
My attempt to fix #224. Sorry about the whitespace, the mix of tabs and spaces drives me nuts in these two files, so I replaced all tabs with 4 spaces. This seems to more or less be the standard in the project.

### DIFF
--- a/Source/Plugins/LfpDisplayNode/LfpDisplayCanvas.cpp
+++ b/Source/Plugins/LfpDisplayNode/LfpDisplayCanvas.cpp
@@ -288,12 +288,9 @@ void LfpDisplayCanvas::update()
         //std::cout << "Updating channel names" << std::endl;
         for (int i = 0; i < nChans; i++)
         {
-
-           String chName = processor->getDataChannel(i)->getName();
-
+            String chName = processor->getDataChannel(i)->getName();
             lfpDisplay->channelInfo[i]->setName(chName);
             lfpDisplay->setEnabledState(isChannelEnabled[i], i);
-
         }
 
         if (nChans == 0) lfpDisplay->setBounds(0, 0, getWidth(), getHeight());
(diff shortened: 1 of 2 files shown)


mspacek:
Here's how the channel info display now looks, at 3 different zoom levels, given the channel map shown. The layout updates immediately whenever channel order or enable/disable change in the channel map.

large
medium
small
chanmap

jsiegle:
Just tested it out, looks great! My only suggestion would be to add a button to toggle between this scheme and the consecutive numbering. Some will prefer to display things the old way. Let me know if that sounds doable, or if you want me to take a stab at it.

mspacek:
Thanks!

I honestly don't see why anyone would ever want it the old way. I'm working on other things right now, such as adding a button to toggle the AUX chans (which I normally don't want to see in any of the blocks), and overall spending too much time diving deep into the code. So I'm not very motivated to spend additional time adding a button to toggle the old numbering behaviour in the LFP Viewer. So, maybe you can take a stab at it if you feel it's important.

aacuevas:
As for why anyone would want the current behavior, we have many users who share the same signal chain for different electrodes, whose mapping configuration just load into the channel mapper to reorder the raw channel configuration into something meaningful for them. For those users it is important that the labels and numbering in the viewer correspond with their probes map instead of the headstage raw ordering.

If you can not add the toggle we'll try to add it before merging this changes. Thanks for the work!

PR #227, opened 2018-06-08
Add AUX chans toggle button to RhythmNode, improve editor layout

mspacek:
* Complements the ADC chans toggle button, and allows for multiples of 64 instead of 67 channels throughout
* Improve consistency of widget sizing and spacing

I haven't thoroughly tested this on the dev branch (I'm mostly working off the master branch), but it seems to work.

### DIFF
--- a/Source/Plugins/RhythmNode/RHD2000Editor.cpp
+++ b/Source/Plugins/RhythmNode/RHD2000Editor.cpp
@@ -239,11 +239,11 @@ void FPGAchannelList::update()
         channelComponents.add(comp);
     }
 
-    Label* lbl = new Label("TTL Events","TTL Events");
+    Label* lbl = new Label("TTL Events", "TTL Events");
     lbl->setEditable(false);
-    lbl->setBounds(numActiveHeadstages*columnWidth,40,columnWidth, 25);
+    lbl->setBounds(numActiveHeadstages*columnWidth, 40, columnWidth, 25);
     lbl->setJustificationType(juce::Justification::centred);
-    lbl->setColour(Label::textColourId,juce::Colours::white);
+    lbl->setColour(Label::textColourId, juce::Colours::white);
     staticLabels.add(lbl);
     addAndMakeVisible(lbl);
 
@@ -556,7 +556,7 @@ RHD2000Editor::RHD2000Editor(GenericProcessor* parentNode,
     impedanceData = new ImpedanceData();
     impedanceData->valid = false;
 
-    // add headstage-specific controls (currently just an enable/disable button)
+    // add headstage-specific controls (currently just a toggle button)
     for (int i = 0; i < 4; i++)
     {
         HeadstageOptionsInterface* hsOptions = new HeadstageOptionsInterface(board, this, i);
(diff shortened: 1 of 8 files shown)


mspacek:
Here's what it looks like. With the AUX chan toggled off, there are 64 chans everywhere instead of 67.

auxchanbutton

aacuevas:
Thanks, this might be useful to reduce signal clutter.
I am unable to test this right now but, from what I've seen, even if aux channels are disabled their related data from the Rhythm data frame is still being copied into the DataBuffer. Perhaps that should be disabled too, since many internals presume that the number of channels in the buffer is the same as the channel count the DataThread provides.

mspacek:
Yes, I think you're correct. At one point, I indeed wasn't copying the aux data in `RHD2000Thread::updateBuffer()` when the aux chans were disabled, but somehow that seemed to create problems, so I removed the `if` clause. I was dealing with some other issue at the same time, and maybe I got confused.

I just noticed `Rhd2000DataBlock::calculateDataBlockSizeInWords()`. Should that calculation exclude the `+3` for the aux chans? Or the `+3+1` including the filler word? `Rhd2000DataBlock::fillFromUsbBuffer()` might need changes too. However, neither of these care if the ADC chans are disabled, so those functions might be too low level for me to worry about.

I should be able to test again in a few days.

aacuevas:
The Rhd2000DataBlocks are indeed too low level to worry about. The Rhythm firmware always sends those three words via USB as part of the packet definition (the chipSettings.enableAuxX() methods only disable the physical pins on the Intan chip), same as ADC data. "Disabling" them from the GUI point of view means just ignoring them and not copying them into the working buffer.

mspacek:
OK, I've overwritten what I previously pushed, 3 separate commits now for clarity. Now when aux chans are disabled, they're no longer copied in `RHD2000Thread::updateBuffer()`

I also added some new comments to `RHD2000Thread::updateBuffer()`. Could you please check that they're accurate?

I'm unclear about this line, [1570 in RHD2000Thread.cpp](https://github.com/open-ephys/plugin-GUI/pull/227/files#diff-07ad4d2eb89a2f3a2cab536ba57b586aR1570), right before the aux chans are copied:
```C++
auxIndex += 2 * numStreams; // mspacek: not clear why this is here
```
When I comment it out, everything seems to run fine, with all combinations of aux and adc toggles. I don't currently have any real signals on either aux or adc chans to test with, so everything looks identical in both cases (50 Hz line noise on the neural channels, positive saturation on the aux chans, 0V on the adc chans. I think).

I'm actually very unclear about the whole block of aux chan copying code that follows. Could someone clarify it for me so I can add some comments to the code?

aacuevas:
Ok, first you need to understand how Rhythm command structure works. The firmware is continuously sending a set of 35 commands to the Intan chips, of those commands, 32 are "CONVERT" commands which sample the amplifier while the other 3 are auxiliary command which are not always the same, but 3 command chains that perform various tasks. Note that depending on the specific command their output might be meaningful or not, but the firmware sends its results nonetheless. One of these commands are auxiliary ADC sampling (the aux channels, not to be confused with aux commands).

In the specific case of how we configure the system, following Intan's example code, we set the command chains to sample the auxiliary inputs, as well as the temperature sensor at 1/4th sampling frequency, cycling within neural samples. (So sample 1 will contain aux 1, sample 2 aux 2, sample 3 aux 3 and sample 4 internal temperature). Command chains 1 and 3 are filled with control commands that return no meaningful value. Since aux data is samples at Fs/4, we opted for a simple upsampling, repeating each value 4 times until a new one is received.

So the line you mention just skips over AuxCmd1, which has no meaningful value, to AuxCmd2 which has the auxiliary channel data.
Following this, we check which aux channel corresponds to the current neural sample, skipping the temperature sensor, and fill it in into an array that will hold them for the next samples. Finally, we simply fill the three aux data channels with the most recent data.

You can find more detailed info in RHD2000DataThread::updateRegisters for command chain creation as well as these documents from Intan:
http://intantech.com/files/Intan_RHD2000_USB_FPGA_interface.pdf
http://intantech.com/files/Intan_RHD2000_series_datasheet.pdf
I hope this clarifies the method for you.

mspacek:
Complicated! Thanks for that. I didn't know temperature was separate from the 3 aux chans, or about the 1/4 sampling. OK, I've replaced that comment.

jvoigts:
@aacuevas that's an awesome overview - maybe it even make sense to link to this from the wiki?

mspacek:
I've been using these changes for a while now in my own branch without issue. Only thing I just noticed is that if the AUX or ADC chans are toggled a few times *after* acquisition has been started and stopped at least once, then eventually it segfaults:
```
...
JUCE Assertion failure in juce_Path.cpp:288
JUCE Assertion failure in juce_Path.cpp:288
JUCE Assertion failure in juce_Path.cpp:267
JUCE Assertion failure in juce_Path.cpp:288
JUCE Assertion failure in juce_Path.cpp:288
JUCE Assertion failure in juce_Path.cpp:288
Segmentation fault (core dumped)
```
However, it turns out this was already the case in the dev branch for the ADC button, so it's not a regression.

PR #237, opened 2018-07-03
Few important fixes related to EvntTrigAvg plugin

melodieborel:


aacuevas:
Hi,

Could you please make different branches for different pull requests, instead of a big one? That way it will be easier to manage the merges. Also, please use the development branch as target.

Regarding the specific commits, those related with sending data back to the Acquisition Board won't be able to be merged as, while they work for a specific configuration of one Rhythm node and one TTL/DAC sink, anything we add into the GUI must support multiple combinations. We are still trying to to design a standard and general way to send data back the chain.

The Event Triggered Averager fixes are very useful. Thanks for those! Just, when trying to discern a rising from a falling pulse, is there a reason you're not using the TTLEvent::getState() method? that way it would be easy to add in the future a button to toggle between rising or falling peak trigger.

Best,
Aarón

PR #238, opened 2018-07-05
Revert change to the recording engine ID

beOn:
I don't know if this was a mistake, or what, but changing this broke some tools of mine.


aacuevas:
One of our collaborators is working with a modified version of the binary format, this string seems to have slipped through his commits. Thanks!

beOn:
Sure thing!

PR #244, opened 2018-07-23
Add float and double cases where missing in InfoObjects

ethanbb:
I noticed that the FLOAT_ARRAY and DOUBLE_ARRAY cases seemed to be missing in `getTypeByteSize` and `setDefaultNameAndDescription` in InfoObjects.cpp. I'm guessing this is just a mistake. Hoping to use a double event channel to send info about stimulation results from one plugin to another down the signal chain.


aacuevas:
Yes, it was just an overlook. Thanks!

PR #260, opened 2018-09-28
PSTH fixes

melodieborel:
Hello,
after struggling with it (I am new with git), here is a minimal pull request concerning 3 PSTH fixes : 
- Rising edge TTL detection
- Display of more than 16 channels
- Sampling rate to acquisition rather than audio.
I hope this kind of format is more convenient.

Best regards


aacuevas:
Yes, it is. Thank you!

PR #262, opened 2018-10-03
Fix bug where splitters don't always update their editors correctly

ethanbb:
Currently, there's a bug where sometimes a splitter editor will have the wrong branch selected until you manually switch it to the other branch and back and it gets in sync with the visible destination processor(s). As far as I can tell, this happens sometimes when loading a signal chain, and when selecting a processor on the processor graph that requires a switch in a splitter that comes before it in the chain in order to become visible. This can be particularly annoying when you have a chain with multiple splitters that you're loading often.

The problem seems to occur when calling `updateVisibleEditors` with action ACTIVATE or UPDATE. In these calls, `Splitter::setPathToProcessor` is used to switch splitters in step 4 if necessary, but this function doesn't actually update the editor, just the splitter's `activePath` and `destNode`. I didn't want to modify this function because it's also used in `ProcessorGraph::updateConnections`, where I'm guessing we might not want to change any of the editors. Instead, I used what's shown here as a workaround. It's pretty indirect though, and it would be good to do a more through refactoring of `Splitter` and `SplitterEditor` since there are a lot of mysterious differences between functions that sound like they should do the same thing.

Also includes some minor cleanup of `Splitter` and `SplitterEditor`.

### DIFF
--- a/Source/Processors/Splitter/Splitter.h
+++ b/Source/Processors/Splitter/Splitter.h
@@ -55,16 +55,17 @@ public:
     /** Nothing happens here, because Splitters are not part of the ProcessorGraph. */
     void process(AudioSampleBuffer& buffer) override {}
 
-    bool isSplitter()
+    bool isSplitter() const override
     {
         return true;
     }
 
-    void switchIO(int);
-    void switchIO();
-    void setSplitterDestNode(GenericProcessor* dn);
+    void switchIO(int) override;
+    void switchIO() override;
+    
+    void setSplitterDestNode(GenericProcessor* dn) override;
 
-    void setPathToProcessor(GenericProcessor* processor);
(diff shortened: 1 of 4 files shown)


aacuevas:
Hi,

We are currently working on revamping the signal chain. That includes splitters and mergers, which will be almost rewritten from scratch, so I'm not sure if we should merge changes to them at this stage, since we intend for the changes to be included for the next release.

Details about that are in #266 which I should had written some time ago to avoid situations like this. sorry about that.

ethanbb:
Yeah, that's OK. I mean I guess it was a bit of a waste of time, but looking through the Splitter and Merger code, as I said I saw there was a lot of inconsistency and it makes sense that you would be rewriting them. So I'm glad that's happening!

Issue #264, opened 2018-10-08
Upgrade ZeroMQ version for Windows


ethanbb:
Hi maintainers,

Using a chain with the Event Broadcaster on Windows, it's been a common occurrence that almost every time we close the program, there's a crash (unhandled exception). After ignoring it for a long time, I finally tracked down the ZMQ debug symbols and looked into it. It turns out that it is a bug with version 4.0.4 of ZMQ that happens only when using it in a DLL, as described here: https://github.com/zeromq/libzmq/issues/1144. It was fixed in version 4.2.0.

Therefore, I suggest we should upgrade the included Windows ZeroMQ library to the current stable release, 4.2.5. I can test it out for a while if needed.

Thanks!

ethanbb:
Actually never mind, the version of the issue we're having still isn't fixed in ZeroMQ, as noted here: https://github.com/zeromq/libzmq/issues/1708. It seems the real solution is to just make sure the ZMQ context is destroyed before the DLL gets unloaded. The easiest way is probably to just have one context per instance of the plugin, rather than making the context static as it is now.

PR #265, opened 2018-10-08
Improve Event Broadcaster robustness and feedback on error

ethanbb:
This incorporates a few changes to try to make the Event Broadcaster more predictable and less frustrating to use:

* Previously, the socket would silently fail to bind to the port every other time you clicked "Restart connection" (at least on my machine), due to the old socket not actually closing and freeing the port immediately. This also caused race conditions when loading a signal chain. The issue is discussed here (for the Node.js version): https://github.com/JustinTulloss/zeromq.node/issues/214. To solve, as suggested there, I added a call to `zmq_unbind` which frees the port immediately. The old socket then only gets closed if the new one is bound successfully; otherwise, it gets rebound to the original port.

* When loading a signal chain, previously the editor would not be updated to show the loaded port. Fixed this by adding a method in the editor to update the displayed port and calling this whenever the port is set.

* If 5557 was already in use, a new Event Broadcaster would default to port 0 (invalid), which is not ideal. Instead, changed the constructor to keep looking for a free port if it receives the `EADDRINUSE` error from `setListeningPort`.

* Made the editor report errors in the status bar if they occur when changing the port or restarting the connection.

I briefly looked at the Network Events plugin as well, but it looks like the code there is substantially different even though the editor's appearance is similar to Event Broadcaster. So I didn't want to get involved with that too given that this is already a medium-sized PR.

### DIFF
--- a/Source/Plugins/EventBroadcaster/EventBroadcaster.cpp
+++ b/Source/Plugins/EventBroadcaster/EventBroadcaster.cpp
@@ -22,6 +22,29 @@ std::shared_ptr<void> EventBroadcaster::getZMQContext() {
     return ctx;
 }
 
+int EventBroadcaster::unbindZMQSocket()
+{
+#ifdef ZEROMQ
+    void* socket = zmqSocket.get();
+    if (socket != nullptr && listeningPort != 0)
+    {
+        return zmq_unbind(socket, getEndpoint(listeningPort).toRawUTF8());
+    }
+#endif
+    return 0;
+}
+
+int EventBroadcaster::rebindZMQSocket()
+{
+#ifdef ZEROMQ
+    void* socket = zmqSocket.get();
+    if (socket != nullptr && listeningPort != 0)
+    {
+        return zmq_bind(socket, getEndpoint(listeningPort).toRawUTF8());
+    }
+#endif
+    return 0;
+}
 
 void EventBroadcaster::closeZMQSocket(void* socket)
 {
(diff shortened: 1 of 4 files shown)


aacuevas:
I was just writing about that.
Adding a reference count is what I was thinking about #264 so perfect.

Thanks!

Source/Plugins/EventBroadcaster/EventBroadcaster.cpp
```diff
@@ -11,17 +11,85 @@
...
+}
+
+// ZMQContext is a ReferenceCountedObject with a pointer in each instance's 
+// socket pointer, so this only happens when the last instance is destroyed.
+EventBroadcaster::ZMQContext::~ZMQContext()
+{
+    ScopedLock lock(sharedContextLock);
+    sharedContext = nullptr;
+#ifdef ZEROMQ
+    zmq_ctx_destroy(context);
```
ethanbb: Maybe would be better to release the lock before calling this (by putting the 2 lines above in their own block). Theoretically all the sockets *should* be gone if this is being called, but it doesn't hurt and since this call blocks if there are still any alive sockets, there could be some potential for a deadlock.

Source/Plugins/EventBroadcaster/EventBroadcaster.cpp
```diff
@@ -11,17 +11,85 @@
...
+#endif
+}
+
+void* EventBroadcaster::ZMQContext::createZMQSocket()
+{
+#ifdef ZEROMQ
+    jassert(context != nullptr);
+    return zmq_socket(context, ZMQ_PUB);
+#endif
```
ethanbb: Realized soon after the last commit that there should be a return statement for the case that ZEROMQ isn't defined (would just return `nullptr`). I have a feeling this would produce at least a warning on Linux.

PR #263, opened 2018-10-08
Compatibility for Visual Studio 2017

melodieborel:
Adds a folder for Visual studio 2017 in 'builts' ; updates the .gitignore accordingly ; and modifies the .vcxproj (<WindowsTargetPlatformVersion>10.0.17134.0</WindowsTargetPlatformVersion>  added to the <PropertyGroup Label="Globals"> ) and all v120 changed for v141

PR #270, opened 2018-11-15
Additional changes for VS2017 compatibility

ethanbb:
This includes the changes in #263 - I tried to make a PR directly to @melodieborel's fork, but it wouldn't let me for some reason. Anyway, here are the additional changes I made:

* Added missing PulsePalOutputCanvas cpp file to project
* Added `.vs` local settings folders (new in 2017) for both solutions to `.gitignore`
* Got rid of the `std::tr1` stuff in both instances of the DSP library. It's deprecated and was causing errors when compiling, and it looks like it was only used in one block of code in the library which was already disabled using an `#ifdef 0`.

Now all 4 configurations compile and seem to work fine for me (on Windows 7). Anybody else want to test it out?


aacuevas:
Merging this is actually far from trivial.
The main issue is that, plugins aside, core GUI build files are not created manually but automatically through projucer, which handles stuff like adding or removing files from the project or the version number.

Thus, a manually created VS2017 project could not properly work the next time we update settings through projucer.

Maybe updating projucer to a newer version that supports VS1027 could solve the issue for the core GUI (the plugin projects in this request would also be needed and much appreciated) but that could come with its own share of issues.

ethanbb:
@aacuevas:

OK, I see your point. Since developing plugins doesn't involve using the Projucer at all, I didn't realize the main app depended on it and was thinking that what it does is simple enough that we could effectively "eject" from the Projucer system and just do everything manually. But it's probably better to keep that framework and maintain the ability to update the Juce version, for instance, and take advantage of current/future improvements to the Juce library.

So to move to VS2017 properly, we'd have to upgrade to Juce 5. As a major version change, I'm sure that would require some fiddling to get it to work properly (and ideally remove things that are deprecated). ROLI also made some questionable changes to the licensing and privacy policy in the new version, and while after some digging it looks like it shouldn't ultimately affect projects that use the GPL, maybe still worth some discussion. Finally this is all probably lower priority than the architectural changes that are underway with Open Ephys itself (signal chain, etc.), so maybe this should wait until that's finished.

All that being said, I think we should eventually have a discussion on a separate issue about upgrading to Juce 5.

jsiegle:
Updating to Juce 5 will likely be the first assignment for the new engineers we're about to hire at the Allen Institute. At that point, we'll be able to add VS2017 compatibility. I agree that the new license is not entirely straightforward, but it doesn't seem like it will affect our use of the library.

ethanbb:
Going to close this since it's out of date. However, it should be possible to update to VS2017 (or 2019!) now that we're using CMake instead of the Projucer, and the changes listed here may still be useful to reference when doing so.

PR #276, opened 2018-12-13
Clean up and improve Network Events

### LINKED ISSUE #174: Network Events causes crashes when restarting connection
Clicking the [restarting connection] button in the Network Events plugin can sometimes cause crashes for the GUI application on the Testing branch. There are 3 behavior's that I've observed so far:

Very rarely, clicking Restart Connection soon after adding the plugin can cause an application crash.
Much more often, clicking Restart Connection more than once (either in rapid succession or with a several second delay) can cause the application to hang.
Sometimes, if the Restart Connection button is clicked and the application survives without crashing, deleting the Network Events plugin can cause a crash.

This sounds like an exception or error from attempting to bind/connect to an unavailable socket.

ethanbb:
I started out just wanting to fix the port on the editor not updating when loading a signal chain. I found some more issues with the plugin though, so I decided to try fixing them. Sorry for the huge diffs... However, I think you'll find it resolves most of the instability issues that Network Events has suffered from. Also fixes #174.

Overview of changes:

* Uses RAII wrappers for the context and socket to avoid leaking memory. In particular, similar to the latest Event Broadcaster, the context is shared between instances via a raw pointer but "owned" jointly by all sockets using reference-counted pointers, so that it is destroyed after all instances of the plugin are destroyed. This avoids a crash with ZeroMQ on Windows that happens when a context is destroyed during unloading of a DLL.

* Avoids destroying the sockets in the main thread (or interacting with them at all except for creating them and passing them to the thread using a lock). ZMQ sockets are not thread-safe, and the call to `zmq_close` on restart, port change, and shutdown was causing some crashes - possibly the cause of #174.

* Doesn't block forever on `zmq_recv`; instead has a timeout of 100 ms. This means possibly some added latency while the thread checks whether it should exit, change port or restart between calls, but these checks should be extremely fast, and the upside is that we don't need to destroy the context each time we need to take one of these actions. The context  should typically persist for the lifetime of the program, and creating a new one has a high overhead. As a consequence:

    * When you have two instances of the plugin and then one is deleted, no longer gets into the unfortunate state where the context has been destroyed and there's no way to make a new one for the remaining instance to use. 

* Updates the port label to always show:
   * The actual bound port if connected
   * The string `<no cxn>` if not connected
   * The string `<no zeromq>` if it was not compiled with ZeroMQ

* Allows inputting a "*" as the port to use any available port. The label then updates to show what port was actually selected. 

* If a port change fails and a socket was connected on the previous port, keeps using that port and socket.

* Sends the status message when receiving an event from the thread rather than the process function, so that it shows up even if acquisition isn't running.

* Simplifies `StringTS` significantly - it doesn't really need to be more complicated than a `String` coupled with a timestamp. In fact, we could use `std::pair` instead, but having names for the members is nice.

* Deletes the code related to simulations, since it seemed to be unused. This is all in the last commit since I wasn't sure if it might still be used at some point (maybe for debugging?), so that can be reverted if necessary.

I realize this might take a while to look through, but I hope you get around to it at some point.

### DIFF
--- a/Source/Plugins/NetworkEvents/NetworkEvents.cpp
+++ b/Source/Plugins/NetworkEvents/NetworkEvents.cpp
@@ -35,67 +35,6 @@ const int MAX_MESSAGE_LENGTH = 64000;
     #include <unistd.h>
 #endif
 
-
-StringTS::StringTS()
-    : timestamp(0)
-{}
-
-
-StringTS::StringTS(String S, int64 ts_software)
-    : str(S)
-    , timestamp(ts_software)
-{}
-
-
-StringTS::StringTS(MidiMessage& event)
-    : timestamp(EventBase::getTimestamp(event))
-{
-    if (Event::getEventType(event) != EventChannel::EventChannelTypes::TEXT)
-    {
-        return; // only handles text events
-    }
-
-    const uint8* dataptr = event.getRawData();
-    // relying on null terminator to get end of string...
-    str = String::fromUTF8(reinterpret_cast<const char*>(dataptr + EVENT_BASE_SIZE));
-}
-
-
-std::vector<String> StringTS::splitString (char sep) const
-{
-    String curr;
-
-    std::list<String> ls;
-    for (int k = 0; k < str.length(); ++k)
-    {
-        if (str[k] != sep)
-        {
-            curr+=str[k];
-        }
-        else
-        {
-            ls.push_back (curr);
-            while (str[k] == sep && k < str.length())
-                ++k;
-
-            curr = "";
-            if (str[k] != sep && k < str.length())
-                curr += str[k];
-        }
-    }
-    if (str.length() > 0)
-    {
-        if (str[str.length() - 1] != sep)
-            ls.push_back (curr);
-    }
-
-    std::vector<String> Svec (ls.begin(), ls.end());
-    return Svec;
-}
-
-
-/*********************************************/
-
 NetworkEvents::NetworkEvents()
     : GenericProcessor  ("Network Events")
     , Thread            ("NetworkThread")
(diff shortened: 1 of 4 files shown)


Source/Plugins/NetworkEvents/NetworkEvents.cpp
```diff
@@ -35,200 +35,75 @@ const int MAX_MESSAGE_LENGTH = 64000;
...
-    // first, close existing thread.
-    closesocket();
-
-    // allow some time for thread to quit
-#ifdef WIN32
-    Sleep (300);
-#else
-    usleep (300 * 1000);
-#endif
+    ScopedPointer<Responder> newResponder = Responder::makeResponder(port);
```
ethanbb: We could make the socket within the thread instead (discarding it without destroying the old one if it fails), and just set urlport here (which we could make atomic) to indicate a port change. Then we wouldn't need `nextResponder` or the lock. I did it this way so that the caller would immediately know whether the change was successful, but that's actually not a big deal since the thread can just update the port on the editor if the switch fails.

Source/Plugins/NetworkEvents/NetworkEvents.cpp
```diff
@@ -35,200 +35,75 @@ const int MAX_MESSAGE_LENGTH = 64000;
...
 {
     setProcessorType (PROCESSOR_TYPE_SOURCE);
 
-    createZmqContext();
-
-    firstTime = true;
-    responder = nullptr;
-    urlport = 5556;
-    threadRunning = false;
+    if (!setNewListeningPort(5556))
```
ethanbb: If we changed `setNewListeningPort` to not create the socket and return feedback right away, we couldn't do this, so maybe it would be better to just choose the port automatically by inputting 0. Of course loading a signal chain would still switch the port to the saved value, if possible.

Source/Plugins/NetworkEvents/NetworkEvents.cpp
```diff
@@ -535,89 +247,100 @@ void NetworkEvents::process (AudioSampleBuffer& buffer)
...
-        // failed to open socket?
-        std::cout << "Failed to open socket: " << zmq_strerror (zmq_errno()) << std::endl;
-        return;
-    }
-
-    threadRunning = true;
-    unsigned char* buffer = new unsigned char[MAX_MESSAGE_LENGTH];
-    int result = -1;
+    ScopedPointer<Responder> responder;
+    char buffer[MAX_MESSAGE_LENGTH];
```
ethanbb: This should probably be a `HeapBlock` rather than an array on the stack, actually. I was thinking I could use a plain array since the size never changes, but I wasn't considering the general rule that large arrays should be on the heap.

ethanbb:
I'm going to go ahead and make another commit with these changes, unless anyone disagrees or wants to discuss them further.

jsiegle:
Works beautifully, thanks again for this!

ethanbb:
You're welcome! I'm glad you didn't have any issues with it.

PR #281, opened 2019-01-02
Event Broadcaster socket reorganization

ethanbb:
This changes the socket in Event Broadcaster from being a derived class of `std::unique_ptr` (something I did before that was weird) to a plain object that does cleanup in its destructor and is held in a `ScopedPointer`. This is more like Network Events and generally more in line with the style of the rest of the codebase. Also changes most actions involving the socket to be methods of the socket object, which I think is cleaner and is also similar to Network Events.

(Sorry to keep revisiting the ZeroMQ plugins...at this point I'm just trying to organize some changes I made a while ago into reasonable-sized PRs.)


aacuevas:
Weirdly enough, this pull request conflicts with your precious one ( #278 ). Could you please solve the conflict? Thanks!

PR #282, opened 2019-01-07
Fix ports sometimes not being unbound in time when loading a signal chain

ethanbb:
This fixes the following bug: 
* You have a signal chain open with one or more Network Events or Event Broadcasters
* You try to load a new signal chain that has one or more Network Events* or Event Broadcasters bound to the same ports as plugins in the previous chain
* Those plugins in the new chain are unable to bind to the correct ports, even though they don't conflict with other plugins in the new chain.

\* I thought I had observed this happening with Network Events previously, but currently it seems to only affect Event Broadcasters.

This means loading a signal chain is sometimes not an idempotent operation (doing it multiple times does not produce the same result every time), which could be confusing.

The reason this happens is obscure. The ports that were bound in the old chain should be available again as soon as the old sockets are unbound, which happens as each processor is destroyed. Each processor is removed from the `ProcessorGraph` strictly before the new ones are created when loading a new signal chain. However, the `removeNode` function in `AudioProcessorGraph` does not necessarily destroy  the processor immediately; it asynchronously calls the function `buildRenderingSequence`, and only once this function returns are all the references to the reference-counted `Node` object that owns the processor instance deleted.

Basically this means that processors are destroyed asynchronously when clearing the chain. (Incidentally, I think this may also explain why loading a signal chain with Rhythm FPGA fails to find the board if there was an instance of the FPGA plugin in the old signal chain.) The solution is to only bind the ports of newly created instances asynchronously. These binding operations should then fall later in the message queue than destroying/unbinding the old instances. I've added `AsyncUpdater` functionality to Network Events and Event Broadcaster so that port changes happen asynchronously during initialization and when loading a signal chain.

### DIFF
--- a/Source/Plugins/EventBroadcaster/EventBroadcaster.cpp
+++ b/Source/Plugins/EventBroadcaster/EventBroadcaster.cpp
@@ -147,19 +147,13 @@ int EventBroadcaster::setListeningPort(int port, bool forceRestart, bool searchF
 {
     if (!synchronous)
     {
-        asyncOptions.port = port;
-        asyncOptions.forceRestart = forceRestart;
-        asyncOptions.searchForPort = searchForPort;
+        const MessageManagerLock mmLock; // since the callback happens in the message thread
         
-        if (asyncOptions.called.exchange(1) == 0)
-        {
-            MessageManager::callAsync([this]
-            {
-                asyncOptions.called = 0;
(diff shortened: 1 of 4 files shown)


ethanbb:
Closing; moved to open-ephys-plugins/ZMQPlugins#3.

PR #290, opened 2019-01-31
Fix only 8 channels saved correctly in first recording

ethanbb:
This fixes a recently-introduced bug where, after starting the GUI, the first recording only saves the first 8 channels correctly; any channels that are selected for recording but are not one of the first 8 _total_ channels are saved as all zeros. After the first recording, if the number of channels sent to the `RecordNode` increases by adding processors to the signal chain, these new higher-numbered channels are also not recorded properly during the first recording when they are in the signal chain.

This happens because the `RecordNode` does not properly set its number of input channels using `setPlayConfigDetails` before the `addConnection` call after the 8th total channel is added (all during `ProcessorGraph::connectProcessorToAudioAndRecordNodes`). This in turn is due to `settings.numInputs` now being initialized to 8, and not increased until the `addSpecialProcessorChannels` call after all the connections are supposed to have been made. The call to `getNextChannel` then erroneously returns -1 after 8 channels have been added.

To fix it, I used `registerProcessor` to update both `settings.numInputs` and the play config details before the channels coming from each processor are added.

### DIFF
--- a/Source/Processors/RecordNode/RecordNode.cpp
+++ b/Source/Processors/RecordNode/RecordNode.cpp
@@ -42,7 +42,6 @@ RecordNode::RecordNode()
     isRecording = false;
 	setFirstBlock = false;
 
-    settings.numInputs = 8;
     settings.numOutputs = 0;
 
     recordingNumber = -1;
@@ -53,8 +52,6 @@ RecordNode::RecordNode()
     hasRecorded = false;
     settingsNeeded = false;
 
-    // 128 inputs, 0 outputs
-    setPlayConfigDetails(getNumInputs(),getNumOutputs(),44100.0,128);
 	m_recordThread = new RecordThread(engineArray);
 	m_dataQueue = new DataQueue(WRITE_BLOCK_LENGTH, DATA_BUFFER_NBLOCKS);
 	m_eventQueue = new EventMsgQueue(EVENT_BUFFER_NEVENTS);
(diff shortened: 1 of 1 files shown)


aacuevas:
Didn't expect this one.
With your changes, does removing the call to setPlayConfigDetails on RecordNode::addSpecialProcessorChannels changes anything? Because if not, we should probably remove it too.

I'm a bit wary about setting the default input count to zero, though. I'll have to check it does not cause issues in some weird configurations (namely, in a chain with no data channels).

jsiegle:
I just encountered the same problem, and this fixes it. However, it only applies to the first sub-processor. Any channels for additional sub-processors are writing all zeros. @aacuevas, can you check if this is the case for you, too? It might just be an issue with the Neuropixels plugin I'm using.

aacuevas:
I've just checked with a tester plugin with 5 32ch streams and it seems to work nicely, all 160 channels recorded.

ethanbb:
@jsiegle That seems odd to me, since as far as I can tell the `ProcessorGraph` functions to link together processors don't do anything special depending on which subprocessor each channel is in - they just connect all the channels output by one to corresponding inputs of another, right? This change also doesn't care about subprocessors; it just uses `settings.numOutputs` to determine the number of inputs, which should be _all_ the outputs regardless of subprocessor. Maybe I missed something though.

 Is it possible that your Neuropixels plugin is setting `numOutputs` incorrectly? If so though, I'd also expect the channels from other subprocessors that are not being recorded to also not be passed down the chain.

jsiegle:
We fixed the bug...it was due to an unrelated issue with sub-processor timestamps.

PR #291, opened 2019-02-15
Save and load all audio settings

ethanbb:
This change allows the audio device type (DirectSound, ASIO, etc.), sample rate, buffer size, and other settings to be saved and loaded to/from signal chain config files. Hopefully this will help prevent mistakes where all the settings from a previous experiment except some important non-default audio settings, and then mysterious performance problems arise (which has happened to me a few times).

Any audio settings that are changed from the defaults should be saved in the `DEVICESETUP` element (generated by `createStateXml`) within the `AUDIO` element. As a backup, the device type, sample rate, and buffer size are also saved as attributes of the `AUDIO` element and restored separately after trying to restore the settings in `DEVICESETUP`.


aacuevas:
Cool!
A couple of questions:
-From what I see, this change would gracefully accept an older save file with just the buffer size, is that right?
-JUCE's documentation does not specify what happens if one tries to set an invalid device type with setCurrentAudioDeviceType. I assume it just choses the default but, have you tested this case?

ethanbb:
- Yup, for an old save file it should just set the buffer size.
- It looks like, in the `initialise` call, if the device type isn't found, it tries to infer the correct device type based on the input and audio device names. If that fails, it just uses the first one in the list. Just tested this quickly and it seems to work. (The `setCurrentAudioDeviceType` call just does nothing if the type isn't found.)

PR #303, opened 2019-06-11
Really view just one subprocessor in the LFP viewer, and also view events not corresponding to any data channels

### LINKED ISSUE #302: Changing how the LFP Viewer deals with subprocessors
To preface, this will all probably be obsolete when we have the streams/probes system (#266), but it's also relevant for what my lab is planning in the near future.

I want to revisit the issue of showing TTL events from Network Events (or other potential sources that provide events with no related continuous data) brought up in PR #277. The problem is that the LFP viewer determines how many samples to draw for each event channel from the samples in its source subprocessor, but these events don't have a real source subprocessor. Instead, currently the Network Events plugin has a dummy subprocessor with no data channels that always provides 0 samples, thus 0 event samples get written.

However, since the viewer is really only supposed to show one subprocessor at a time, it should be unambiguous how many samples are in the current buffer. So by making that actually be the case (by selecting source ID and subproc ID rather than just the latter in the LFP viewer combobox), we can save the full source subproc ID that is being viewed and use that as a default.

The downside is that you wouldn't be able to view data from multiple sources in one LFP viewer simultaneously. I'm actually impressed that it doesn't seem to crash when doing that currently, although if the sample rates are different, of course it doesn't display correctly since we're not doing resampling. Anyway, if there's a "zipper" module coming up in the new system, my understanding is that would make this possible again as long as the sources have the same sample rates.

My question is, does this sound like an improvement that would be good to submit for a PR to `development`, or would it break things other people are doing too much and we should just keep it on our fork?

ethanbb:
This resolves #302. To summarize, the LFP viewer currently has the weird behavior of showing data from all subprocessors with a given index in their source nodes, rather than just one single subprocessor. This is convenient and works sometimes, but the incoming data from different sources could have different sample rates or different amounts of samples per buffer even with the same sample rate; dealing with this is complex and currently causes display glitches in some cases, and maybe instability.

This changes the combobox to let you choose a specific source node/subprocessor combination - for example, "File Reader 111/0" is the first subprocessor of the File Reader with node ID 111. Because there is now only one subprocessor being drawn, we only have to keep track of one current sample rate rather than an array of one per channel; however, I also added maps from subprocessor ID to sample rate and # of channels in the node, so that switching from one subproc to another works correctly.

As a side effect, we can now easily fix the issue brought up in #277 of Network Events TTLs not showing up, because we can use the information (i.e. # of samples) from the subprocessor that is being drawn to infer how to draw events with no associated continuous data. A simple way to do this is to treat any event channels with a sample rate of 0 as "standalone" events not corresponding to continuous data. I'm also making open-ephys-plugins/ZMQPlugins#2 that changes the Network Events event channels to have sample rate 0 so that this will work.

There are a bunch of changes, including some style things I did just to get rid of warnings in the process of working through the LFP Viewer code, so definitely ask if some parts don't look right/make sense!

### DIFF
--- a/Plugins/LfpDisplayNode/LfpDisplayCanvas.cpp
+++ b/Plugins/LfpDisplayNode/LfpDisplayCanvas.cpp
@@ -212,67 +212,34 @@ void LfpDisplayCanvas::endAnimation()
 
 void LfpDisplayCanvas::update()
 {
-	if (true)
-	{
     nChans = jmax(processor->getNumSubprocessorChannels(), 0);
 
-	std::cout << "Num chans: " << nChans << std::endl;
+    std::cout << "Num chans: " << nChans << std::endl;
 
     resizeSamplesPerPixelBuffer(nChans);
 
-    sampleRate.clear();
-    screenBufferIndex.clear();
-    lastScreenBufferIndex.clear();
-    displayBufferIndex.clear();
+    sampleRate = 30000; // default
+
+    for (auto* arr : { &screenBufferIndex, &lastScreenBufferIndex, &displayBufferIndex })
+    {
+        arr->clearQuick();
+        arr->insertMultiple(0, 0, nChans + 1); // extra channel for events
+    }
     
     options->setEnabled(nChans != 0);
     // must manually ensure that overlapSelection propagates up to canvas
     channelOverlapFactor = options->selectedOverlapValue.getFloatValue();
 
 	std::cout << "Checking channels: " << nChans << std::endl;
-	for (int i = 0; i < processor->getNumInputs() + 1; i++) // extra channel for events
+
+    for (int i = 0, nInputs = processor->getNumInputs(); i < nInputs; i++)
     {
-		//std::cout << i << std::endl;
-		if (processor->getNumInputs() > 0)
-		{
-			if (i < processor->getNumInputs())
-			{
-				if (processor->getDataChannel(i)->getSubProcessorIdx() == drawableSubprocessor)
-				{
-					sampleRate.add(processor->getDataChannel(i)->getSampleRate());
-					//std::cout << "Adding sample rate " << processor->getDataChannel(i)->getSampleRate() << std::endl;
-				}
-					
-			}
-			else
-			{
-				//Since for now the canvas only supports one event channel, find the first TTL one and use that as sampleRate.
-				//This is a bit hackish and should be fixed for proper multi-ttl-channel support
-
-				for (int c = 0; c < processor->getTotalEventChannels(); c++)
-				{
-					if (processor->getEventChannel(c)->getChannelType() == EventChannel::TTL)
-					{
-						sampleRate.add(processor->getEventChannel(c)->getSampleRate());
-						std::cout << "Sample rate = " << processor->getEventChannel(c)->getSampleRate() << std::endl;
-
-					}
-				}
-			}
-		}
-		else
+        if (processor->getDataSubprocId(i) == drawableSubprocessor)
         {
-			sampleRate.add(30000);
+            sampleRate = processor->getDataChannel(i)->getSampleRate();
+            break;
         }
-        
-       // std::cout << "Sample rate for ch " << i << " = " << sampleRate[i] << std::endl; 
-        displayBufferIndex.add(0);
-        screenBufferIndex.add(0);
-        lastScreenBufferIndex.add(0);
     }
-
-	lfpDisplay->setDisplayedSampleRate(sampleRate[0]); // only one sample rate possible for now
-	std::cout << "Setting display sample rate to " << sampleRate[0] << std::endl;
 
 	std::cout << "Checking channel alignment: " << nChans << std::endl;
     if (nChans != lfpDisplay->getNumChannels())
@@ -319,8 +286,6 @@ void LfpDisplayCanvas::update()
             lfpDisplay->rebuildDrawableChannelsList();
         }
     }
-    
-	}
 }
 
 
@@ -410,7 +375,7 @@ void LfpDisplayCanvas::updateScreenBuffer()
 			//     std::cout << channel << " " << sbi << " " << dbi << " " << nSamples << std::endl;
 
 
-			float ratio = sampleRate[channel] * timebase / float(getWidth() - leftmargin - scrollBarThickness); // samples / pixel
+			float ratio = sampleRate * timebase / float(getWidth() - leftmargin - scrollBarThickness); // samples / pixel
 			// this number is crucial: converting from samples to values (in px) for the screen buffer
 			int valuesNeeded = (int) float(nSamples) / ratio; // N pixels needed for this update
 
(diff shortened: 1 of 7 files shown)


aacuevas:
Thanks for this!

The only thing I'm not convinced is using a 0 sample rate as indicator that the event is "standalone"
For this kind of events, the GUI sets timestamps based on a reference processor (by default the first continuous data source), using interpolation to approximate the moment of event generation (for example a command received by network) to their matching timestamp.
The same way, the event "sample rate" is set to that of the selected reference processor. This way, getting the timestamp is seconds is as easy as dividing timestamp by sample rate, and everything is contained inside the event info structure.

Using a zero-timestamp as a marker not only feels hackish, it also prevents having this information stored anywhere.

I'd advocate using a different approach. Maybe adding some fields to the Event class itself to register if the source processor had associated continuous data or not. Although I'm generally against adding stuff to the base classes, this could justify it.

ethanbb:
@aacuevas OK, good point. I didn't realize the sample rate on event channels was used for that purpose.

What I had previously, before changing it at the last minute to setting the sample rate to 0 since I thought it was simpler, was that event channels with no source data would set all fields of the metadata `source.channel.identifier.full` to 0 (so that `getChannelSourceID()` would return 0). Maybe that's less hacky? Although the downside is it would be pretty tricky to discover for a new plugin author unless it was explicitly documented.

aacuevas:
I was thinking on adding a field in Event telling if the source processor has continuous data or not.
A simple Event::sourceGeneratesTimestamps() (set to the source processor generatesTimestamps() ) could suffice.

It could also be useful to add some
Event::getTimestampSourceID() and Event::GetTimestampSourceSubprocessor()
methods, to get the exact source the timestamps are coming from.

ethanbb:
OK, I'm not sure I see the full picture yet but it makes sense to me to make this information available in a less hacky way than in the metadata field. So do you want to just leave the PR open for now pending base class changes?

I'm going to revert the last commit that added the check for sample rate = 0, since it seems like we agree that was a bad idea.

ethanbb:
I guess when you create the `Event` object, somehow the constructor would need to know the source node/subproc of the continuous data as well as of the event if those differ? (Instead of using the hacky metadata field.) So you could provide those as optional arguments, which would default to using the same as the event source which is known to the `EventChannel`? Except if the `EventChannel` source doesn't generate timestamps, in which case it would fall back to the current global timestamp source?

A little convoluted, but it would be flexible in that changes to the timestamp source or to the data channel backing an event channel would be reflected in any subsequently created events without making a new `EventChannel` object or needing to stop acquisition.

aacuevas:
I left this for a while, sorry, but I believe it is time to revive it. I would want these changes to get into the next release I am preparing.
I updated a bit the EventChannel class from this discussion. There is a getTimestampsOrigin() method thar returns three possible values:
- timestampsFromContinuousSource, for those generated by a regular source node (like the Rhythm node)
- timestampsFromGlobalSource, for those generated by sources that do not generate timestamps (like the Network Events)
- timestampsDerivedFromChannel, for those that are created by a filter from a continuous channel (like the phase detector)
For this last kind, there is also a dedicated constructor which, instead of a sampleRate has a DataChannel* as argument.
I have also added two methods to get the sourceid and subprocessor which originates the timestamps (i.e., those of the channel). I am not sure if I should add also a channel idx field, so the metadata field would be redundant.

Could you please check the changes and update this PR accordingly if needed?

ethanbb:
This is great, thanks! Working on updating the PR.

Is it me or are the new getter functions not yet implemented? I can't see them in `InfoObjects.cpp`.

ethanbb:
@aacuevas: Should we still keep a separate TTL state for each subprocessor? As in, for this line:

`ttlState[eventSourceNodeId] |= (1LL << eventChannel)`

I'm not sure what should happen if the event is linked to the global timestamp source and not any specific data channel/subprocessor. It seems like having just one global TTL state might be less confusing. Is there a use case for maintaining separate channel namespaces/states for different data sources?

aacuevas:
Yes, the method bodies were missing from the commits, my bad. They are on the repo now.

Regarding your second question, the truth is that the LFP viewers still rely on old core, before event channel objects were a thing. That is also why they only support 8 TTLs with no distinction of source.
I did some additions in preparation for eventual improvements that made such distinctions, but they haven't happened yet.

ethanbb:
OK, so I guess TTL events not linked to a specific data channel could alter the TTL state of all sources? Does that seem logical?

aacuevas:
Do you mean throughout the system or just for the LFP viewer? For the whole system we have to take into considerations the limitations of the JUCE library, as well as make it general enough. "ttl states" is only relevant to visualization, as TTL events have value on their own. Mixing them for visualization is actually a non desirable side effect of the viewers being based on an old architecture. Ideally we should be able to select which TTL events to visualize. But that is an extensive patch outside of the scope of this PR.

For the sake of getting this PR working, I have added a new `CoreServices` method, 'getGlobalTimestampSourceFullId()`, which I believe has enough merit to exist. I have also updated the event channel objects so the timestamp origin is set to that same value when `timestampFromGlobalSource`.

Since for visualization we are, indeed, mixing all ttl events with a shared timestamp source into a single state, for events with `timestampsFromGlobalSource` we can now just use the same method for all three types and set the joint TTL state for each source.

I would like this PR to be integrated into development as soon as possible, as we are planning a new release soon and I feel this should be into it.

Thanks for all the help!

aacuevas:
I do believe there is amisconception with ttls and data links.
Most of the TTLs are actually not data linked. Any TTL inputted into the acquisition board, for example, is not inherently linked to any data. It is just that the timestamps are generated by the acquisition board.
An event generated by a filter is usually linked to a channel. But keep in mind that a filter might generate other kind of events. There could be, for example, a processor that detected peaks and generated a binary event with the amplitude for any local peak. TTLs are just one particular type of event. Of course, those events timings are related to a channel and share timestamps with the processor which created the channels.

And now we get to the third type. Stuff like the network processor creates events (which may be TTL, text messages or binary events) with no relation to any source whatsoever. But when recording (which is the priority) we need to be aware of when they were generated. That is where the global timestamps come into action. Global timestamps are actually "timestamps for processors which do not produce their own but produce events", nothing more.

By default they are interpolated from the first continuous source, but it can be selected with Edit/Timestamp source.

For recording, there no issue at all. Everything get recorded with their timestamps that can, in turn, be converted to seconds, synchronized with other sources, etc... Keep in mind that the priority is always recording, and having all the data present in the recordings.
Visualizing is another issue. For example, one might want to see events in the LFP viewer generated by the network events module, even when they are not related to the continuous channels being shown.

Now, the LFP viewer on itself is in an incomplete state related to that. Right now it just lets you see "8 channels" which originally corresponded to the hardware TTLs. In a ideal scenario, there would be 8 (or N) visualization "channels" whose actual event source could be selected. 
Also, the way it works now, trying to synchronize events and data coming at different rates with different timestamps does not work properly. That why in the LFP viewer we have now we added the selector, so it only shows continuous data related to a single source. Ideally, we should maybe do that the visualized events are also those whose timestamps are related to that source, that should not be too difficult, but is still some work outside the scope of this PR.

ethanbb:
OK, should be working now. See also open-ephys-plugins/ZMQPlugins#4.

aacuevas:
Great, thanks!

PR #304, opened 2019-06-14
Fix make -f Makefile.plugins clean

ethanbb:
This fixes `make -f Makefile.plugins clean` on Linux, which was deleting the whole `build` folder including the bitfiles and `libokFrontPanel.so` library. The issue came up since there are now no common libraries, so `$(addsuffix .so,$(notdir $(COMMONDIRS)))` was an empty list, resulting in removing just `$(BINDIR)`.

I changed it to use `$(addprefix ...)` to prevent this, plus use `$(SHAREDDIR)` rather than `$(BINDIR)` since that's where the libraries are now located.


ethanbb:
Closing since this is irrelevant with the new CMake system.

PR #317, opened 2019-09-15
Prevent crash when recording again after recording session with thread turned off

ethanbb:
The following sequence of actions currently crashes the GUI:

* Turn the record thread off in the recording options
* Make a recording (of course, the data doesn't get saved through the record thread, but maybe some plugin makes its own recording somehow)
* Turn the record thread on again
* Try to make a recording.

This happens because events and spikes are still getting added to the record thread's queues when it's not running, which are never being read. Then, when the queues get cleared when starting the real recording, the `AbstractFifo` doesn't get properly reset, so it reports that there are events and spikes available that have actually been cleared, resulting in a null pointer dereference.

It seems to me that both behaviors are wrong, i.e. the events and spikes shouldn't be added when the record thread is off, and also clearing the queues should reset the `AbstractFifo`. So that's what this PR does.

### DIFF
--- a/Source/Processors/RecordNode/EventQueue.h
+++ b/Source/Processors/RecordNode/EventQueue.h
@@ -77,14 +77,14 @@ public:
 	void reset()
 	{
 		m_data.clear();
-        m_fifo.reset();
+		m_fifo.reset();
 		m_data.resize(m_fifo.getTotalSize());
 	}
 
 	void resize(int size)
 	{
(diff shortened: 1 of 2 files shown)


aacuevas:
Thanks!
For the record, disabling the record thread can be useful for the neuropixel plugin, which can make its own recording via the neurpixel API. Is a bit of a temporary workaround.

PR #320, opened 2019-09-27
Show alert window to confirm before quitting GUI

### LINKED ISSUE #313: Gui can be closed without warning
Accidentally clicking the close X promptly closes the gui even during recording. Was thinking there should probably be a warning box to confirm that closing the gui is desired.

anjaldoshi:
- Shows an alert window that asks for confirmation from the user before closing the GUI. This alert window gets triggered by clicking on the "close" button of the main GUI window, or by pressing OS specific key shortcuts for closing a window (ALT+F4 on Windows & Linux, Cmd+Q on Mac). 

- The user can confirm it by clicking _Yes_ (or Enter key), or deny by clicking _No_ (or Esc key) on the alert window.

- Resolves #313 specified by @greggheller

### DIFF
--- a/Source/Main.cpp
+++ b/Source/Main.cpp
@@ -105,6 +105,7 @@ public:
     //==============================================================================
     void systemRequestedQuit()
     {
+		mainWindow->shutDownGUI();
         //std::cout << "Quit requested" << std::endl;
         quit();
     }
--- a/Source/MainWindow.cpp
+++ b/Source/MainWindow.cpp
@@ -125,15 +125,17 @@ MainWindow::~MainWindow()
 
 void MainWindow::closeButtonPressed()
 {
-	if (audioComponent->callbacksAreActive())
+	if(AlertWindow::showOkCancelBox(AlertWindow::WarningIcon, "Quit Open-Ephys", "Are you sure you want to quit the GUI?", "Yes", "No"))
 	{
-		audioComponent->endCallbacks();
+		if (audioComponent->callbacksAreActive())
+		{
+			audioComponent->endCallbacks();
+		}
+
+		processorGraph->disableProcessors();
+
+		JUCEApplication::getInstance()->systemRequestedQuit();
 	}
-
-	processorGraph->disableProcessors();
-
-	JUCEApplication::getInstance()->systemRequestedQuit();
-
 }
 
 void MainWindow::saveWindowBounds()
--- a/Source/MainWindow.cpp
+++ b/Source/MainWindow.cpp
@@ -125,17 +125,19 @@ MainWindow::~MainWindow()
 
 void MainWindow::closeButtonPressed()
 {
-	if(AlertWindow::showOkCancelBox(AlertWindow::WarningIcon, "Quit Open-Ephys", "Are you sure you want to quit the GUI?", "Yes", "No"))
+
+	JUCEApplication::getInstance()->systemRequestedQuit();
+
+}
+
+void MainWindow::shutDownGUI()
+{
+	if (audioComponent->callbacksAreActive())
 	{
-		if (audioComponent->callbacksAreActive())
-		{
-			audioComponent->endCallbacks();
-		}
-
-		processorGraph->disableProcessors();
-
-		JUCEApplication::getInstance()->systemRequestedQuit();
+		audioComponent->endCallbacks();
 	}
+
+	processorGraph->disableProcessors();
 }
 
 void MainWindow::saveWindowBounds()
--- a/Source/MainWindow.h
+++ b/Source/MainWindow.h
@@ -62,6 +62,8 @@ public:
     /** Determines whether the last used configuration reloads upon startup. */
     bool shouldReloadOnStartup;
 
+	void shutDownGUI();
+
 private:
 
     /** Saves the MainWindow's boundaries into the file "windowState.xml", located in the directory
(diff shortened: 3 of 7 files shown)


jsiegle:
Does this alert window always show up when closing the GUI, or only when acquisition is active? I think it should only appear in the latter case.

anjaldoshi:
> Does this alert window always show up when closing the GUI, or only when acquisition is active? I think it should only appear in the latter case.

I was planning to do it only during acquisition (as advised by @medengineer), but after talking to Gregg, he advised that it should pop-up in any case. 

I can change it to show-up only when data acquisition is active.

aacuevas:
Nice! I was meaning to do exactly this some time.
The only thing to note is that the alert window won't show when shutdown is initiated by other means than the window close button. That is, closing the GUI from the menu or trying to shut down windows with the GUI active will not trigger the dialog,

That might be desired, I am a bit on the fence about that. If we want all cases to trigger from any case I'd move everything but the `systemRequestedToQuit()` line to a new bool function, like `bool MainWindow::shutDownGUI()`, make that function do all the required shutdowns and return true if the dialog was answered positively, return false otherwise and then in Main.cpp `OpenEphysApplication::systemRequestedQuit()` call this function and only run `quit()` if true.

Actually, we should do all the quits this ways, independently of the dialog. Some shutdowns were not being performed correctly like it is now. So this could be a good opportunity to fix it. 

Regarding asking only when acquiring or in any case, I am actually more inclined to the latter. It might be frustrating to set up a complex signal chain and then accidentally closing the GUI before saving it. Ideally we could have a system in place that only asks when not acquiring if there are unsaved changes to the chain, but that requires an infrastructure that is far beyond the scope of this PR, so I would just leave the popup.

greggheller:
Like Anjal mentioned, I’m in favor of asking all the time, not just when acquiring. Hitting enter seems a small price to pay vs the risk of losing something, or even the time it would take to open it back up if you accidentally closed it right before you needed to record. I feel like most users don’t need to close the GUI very often, if your really worried about the prompt getting annoying for users you could add a menu option to toggle between always ask, and only during acquisition (or only during recording? Or never?). 

As far as whether it should ask when shutting down windows os, I don’t have a strong opinion but can offer insight from our use case. The program we use to share a mouse between computers doesn’t work If shutdown is halted by open applications, so we have to physically move the mouse to that machine to continue shutdown which is a (relatively small) pain. I’m not sure how many other users would have this issue but it’s worth considering. It seems a little far fetched that someone would initiate shutdown while recording but maybe prompt then? Maybe also when there are unsaved changes to the chain as suggested above (if you move forward with that feature)? Even better might be to save the unsaved changes as an alternate configuration and ask which to reload when it’s opened again

anjaldoshi:
Me, @jsiegle , and @medengineer were discussing about different cases where such an alert should pop up. We narrowed it down to three cases:

1. If data acquisition is _active_, an alert window should pop-up informing the user about active data acquisition and ask for confirmation before closing the GUI,
2. If _Reload on startup_ is not selected in the file menu, and if there is a signal chain set-up, then the alert window should pop-up with a message reminding the user about unsaved signal chain.
3. A new file menu option (Ask Before Quitting?) should be added which, when selected, should show the alert window every time, no matter the case (except OS shutdown).

These three cases cover most of our scenarios where a user may want the alert window to show up. The only case I'm on the fence about is showing the alert window when an OS shutdown event occurs. This can be a bit risky when the GUI becomes unresponsive due to some reason and the user is trying to restart the computer. Due to the nature of the alert window, it won't allow the OS to shutdown/restart without confirming to quit the GUI on the alert window. Such a case may be rare, but we cannot ignore it. This is just my opinion, so I would like to hear some feedback on this. 

> If we want all cases to trigger from any case I'd move everything but the `systemRequestedToQuit()` line to a new bool function, like `bool MainWindow::shutDownGUI()`, make that function do all the required shutdowns and return true if the dialog was answered positively, return false otherwise and then in Main.cpp `OpenEphysApplication::systemRequestedQuit()` call this function and only run `quit()` if true.

This seems to be a great approach to deal with alert windows, and GUI shutdown in general. I'll definitely incorporate this once we decide on which cases should be feasible for the alert window to pop-up.

aacuevas:
Some notes on the quit methods
JUCE's recommended place to handle anything related to the application quitting (shutting down hardware, resetting stuff and asking for confirmation) is `systemRequestedQuit()`, as it gets called by every system request to gracefully exit the application, while `MainWindow::closeButtonPressed()`only gets called when the close button is pressed, but not by any other means.
This means that the way we are handling shutdown now (in 'closeButtonPressed()`does not turn acquisition off nor does any of the required cleanup when the GUI is closed by other means, the most  notable example being the Quit option in the GUI's own menu.

What the app does not have is a way to discern what action exactly requested the quit. This includes shutdown, as Windows always tries to first close gracefully any application. In this case it would happen the same as if trying to shutdown the machine with an unsaved notepad opened. Windows itself intentionally does not provide any way to get this info, as far as I know, as it treats every application equally. As for the usefulness of this, it is the same as with any other windows application, just preventing an accidental shutdown while work is being done, but at least on Windows 10 it has a timeout that will kill any application that refuses to close by itself if the user does not cancel the shutdown.

Of course, the dialog and everything else gets ignored by any action that forcefully closes the process, such as the task manager, windows' "unresponsive application" dialog, selecting "shutdown anyways" on shutdown or (interestingly enough) closing the debug console. Actions like those just kills any running process.

aacuevas:
A little update,

I have actually moved the shutdown code (without any alert or notification, just what we had) to `systemRequestedQuit()` because some behaviors we had now such as not stopping acquisition when quitting through the menu were definitely bugs.

PR #331, opened 2019-12-23
CAR fix because of xml being loaded twice

markschatza:
Found a bug in how OE handles loading in .xml files. It seems since the upgrade to cmake, xml files are loaded in **twice**! Specifically in CAR this caused channels to get added to the reference list multiple times. The referenced average was then being subtracted from the channels twice. 

This PR is a quick fix that only adds channels to the list if not already present. I wanted to get this PR out ASAP as I'm sure others have been affected by this. A broader look should be had at other plugins that might be affected, or if the double load can/should be removed altogether.

Mark

### DIFF
--- a/Source/Processors/GenericProcessor/GenericProcessor.cpp
+++ b/Source/Processors/GenericProcessor/GenericProcessor.cpp
@@ -1038,38 +1038,38 @@ void GenericProcessor::loadFromXml()
 	update(); // make sure settings are updated
 	if (parametersAsXml != nullptr)
 	{
-		if (!m_isParamsWereLoaded)
-		{
-			std::cout << "Loading parameters for " << m_name << std::endl;
+        if (!m_isParamsWereLoaded)
+        {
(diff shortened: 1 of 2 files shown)


Source/Processors/GenericProcessor/GenericProcessor.cpp
```diff
@@ -1038,38 +1038,38 @@ void GenericProcessor::loadFromXml()
...
+            // load editor parameters
+            forEachXmlChildElement(*parametersAsXml, xmlNode)
+            {
+                if (xmlNode->hasTagName("EDITOR"))
+                {
+                    getEditor()->loadEditorParameters(xmlNode);
+                }
+            }
 
+            forEachXmlChildElement(*parametersAsXml, xmlNode)
```
markschatza: Tabbed this for loop into the if statement. Not sure if this was a bug or feature, but I believe these lines should be in the ```if (!m_isParamsWereLoaded)``` statement.

PR #335, opened 2020-01-29
Lfpviewer beta mutex bug fix

wagenadl:
Improved the code in the beta version of the LFP viewer that generates a AudioSampleBuffer channel for events so there is less redundancy in filling the buffer. 

Much more importantly, fixed longstanding (?) bug that caused timing mismatch between display of traces and events. It was due to not locking the mutex while filling the events channel.

Also fixed minor bug in the Canvas code for the beta viewer: it looped beyond the size of the input array, thus adding a spurious entry to that array.

Apologies for the whitespace changes. I replaced some tabs with four spaces so I can edit in emacs.


wagenadl:
I am closing this pull request because I noticed that this code has already been pulled into "testing." I now realized that I should have created a specific branch for each pull request and will do so in the future.

PR #337, opened 2020-01-30
Event triggered display

wagenadl:
This patch series adds a Combobox to the "editor" for the LFP Viewer Beta that allows the user to select a single event channel as a "trigger source." When the trigger source is "none" (the default), the viewer behaves as before. In contrast, when the trigger source is 1...8, traces are rendered only when a corresponding event is received, exactly analogously to the "trigger" function on an oscilloscope. For now, the event is always aligned with the left edge of the display. I am working on a further patch to lift that restriction and would welcome suggestions on where to squish the control for that function in the UI. Would it be acceptable to shift the Pause button even farther to the right?


medengineer:
Looks good and very useful feature! Shifting the pause button over is not a problem. I'll leave this PR open until the final patch is ready.

aacuevas:
As a suggestion, instead of using 8 TTL channels, would it be possible to list the actual event sources? You could look at the record controller or the evt. triggered averager as examples to see how the list is populated and processed.

The LFP viewer having just toggles for 8 channels and triggering based on channel number alone is a very old part of the viewer which we have not just got to update yet, but it would be interesting if new additions were geared towards that.

wagenadl:
I really like that suggestion, because I agree the numeric event sources are less than elegant. I will look into the implementation details.

- Daniel

On 1/31/2020 5:19 AM, Aarón Cuevas López wrote:

As a suggestion, instead of using 8 TTL channels, would it be possible to list the actual event sources? You could look at the record controller or the evt. triggered averager as examples to see how the list is populated and processed.

The LFP viewer having just toggles for 8 channels and triggering based on channel number alone is a very old part of the viewer which we have not just got to update yet, but it would be interesting if new additions were geared towards that.

markschatza:
Hey this is really cool! I'm not sure exactly what your specific use case for this feature is, but I just pushed an early beta version of a [real-time evoked potential plugin](https://github.com/tne-lab/real-time-ERP). I never thought of using the backbone of LFP viewer and editing it into a new plugin. But I may do that instead of the hacky visualization I have right now. If you're trying to do something similar it may make sense to collaborate on building some sort of "Event Triggered LFP Viewer" plugin.

wagenadl:
For us, the major use case is to look at stimulus artifacts. But I think
it is generally useful to lock display on stimuli. I am currently
working with the core team to integrate the work into the regular "LFP
Viewer" so that the feature can be used together with all of that
plugins features, including spike display, etc.

- Daniel

anjaldoshi:
Closing because a new PR #344 (merged) includes all the commits from this one.

PR #341, opened 2020-03-02
Created a simple UDP server for receiving raw matrix data

jonnew:
- This can be used for communication with e.g. bonsai
- It is very brittle and could use a lot of generalization and
improvment

### DIFF
--- a/Plugins/EphysSocket/CMakeLists.txt
+++ b/Plugins/EphysSocket/CMakeLists.txt
@@ -1,0 +1,16 @@
+#plugin build file
+cmake_minimum_required(VERSION 3.5.0)
+
+#include common rules
+include(../PluginRules.cmake)
+
+#add sources, not including OpenEphysLib.cpp
+add_sources(${PLUGIN_NAME}
+	EphysSocket.cpp
+	EphysSocket.h
+	EphysSocketEditor.cpp
+	EphysSocketEditor.h
+	)
+
+#optional: create IDE groups
+plugin_create_filters()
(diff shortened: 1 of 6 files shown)


medengineer:
Awesome, merging into dev so we can polish it up a bit before sending to master.

jonnew:
Great -- there are lots of things wrong with this right now. The big one is that it assumes input type. Instead, this should be a dropdown option (uint8, int8, ...., F64) rather than just assuming the node is receiving uint16_ts

jonnew:
Also, protection for values in textboxes (they should all be positive) and maximum read size. This is using ipv4 whose max payload is 65 kb. There is no checking to make sure the buffer is smaller than this.

jonnew:
Is there any progress on getting this into the main branch so that it is eventually shipped with the pre-compiled software? It is going to be our link to use the OE GUI to acquire data from our next generation hardware, so its fairly important.

jsiegle:
It now makes more sense to distribute this through the Plugin Installer (available in GUI v0.5.0), so it can be released/upgraded independently of the host application. This is how we want to release all plugins moving forward.

This will involve creating a separate repo, similar to [this one](https://github.com/open-ephys-plugins/phase-calculator). If you want to be included in the commit history, you can create a new repo under your own account with just the "Source" directory, which can be forked to the `open-ephys-plugins` account. Otherwise, we can set everything up from scratch. We will get this sorted out soon as the new release is out, hopefully later this week.

jonnew:
OK, makes a lot of sense. I will create a little repo then.

jonnew:
Actually, until the documentation is up for the plugin creation process, I'd rather not wade into creating one. Go ahead and just use the code to make the plugin

PR #345, opened 2020-03-19
Event triggers in lfpdisplaynode

wagenadl:
This is the same feature I previously implemented for the LFPViewerNodeBeta, now in LFPViewerNode itself. User can select a trigger source to align display on events, just like on an oscilloscope. This is very useful in combination with TTL inputs on the "Rhythm FPGA." It can also be used fruitfully with my "FutureThreshold" plugin (https://github.com/wagenadl/open-ephys-plugins).

The "trigger" feature works with old-fashioned "event numbers" (1--8) rather than more modern "event sources." This can presumably be changed, but I do not yet have an understanding of the modern system.

I also added some shorter time scale choices to the "Timebase" combobox. These are useful for looking at fast events surrounding triggers.

Note that it is intentional that the trigger event is positioned at 1/3 of the display width; that way, some pre-trigger data is visible as well.

### DIFF
--- a/Plugins/LfpDisplayNode/LfpDisplayNode.cpp
+++ b/Plugins/LfpDisplayNode/LfpDisplayNode.cpp
@@ -43,8 +43,10 @@ LfpDisplayNode::LfpDisplayNode()
         arrayOfOnes[n] = 1;
     }
 
-	subprocessorToDraw = 0;
-	numSubprocessors = -1;
+    subprocessorToDraw = 0;
+    numSubprocessors = -1;
+    triggerSource = -1;
+    latestTrigger = -1;
 }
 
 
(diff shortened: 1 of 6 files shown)


Plugins/LfpDisplayNode/LfpDisplayNode.cpp
```diff
@@ -353,62 +359,68 @@ void LfpDisplayNode::handleEvent(const EventChannel* eventInfo, const MidiMessag
...
-		else
-		{
-			newIdx = nSamples - samplesLeft;
-		}
+        displayBufferIndices[i][chan] = newIdx;
+        if (latestCurrentTrigger >= 0)
+        {
+            latestTrigger = latestCurrentTrigger + index;
+        }
+    }
```
anjaldoshi: A major difference between "beta" LFPViewer and "main" LFPViewer is that the main one supports switching between different sub-processors while data acquisition is active. This means that, when there are multiple inputs connected to the main LFPViewer, one can switch between these inputs on the fly by selecting any one from the drop-down menu of LFPViewer in the signal chain. To provide this functionality, instead of having a single display buffer, LfpDisplayNode now manages multiple display buffers, one for each input sub-processor. 

Considering all of this, the way you update the `lastTrigger` value in the `finalizeEventChannhels()` method should be different as compared to the beta version. As you can see in the method definition, the main version has a for loop that iterates through each input sub-processor, and then updates the `displayBufferIndex` for that sub-processor's event channel with new index value. But, we don't need to update the `lastTrigger` value for each sub-processor, because if we do that for all sub-processors, then the `lastTrigger` value is always going to use the `displayBufferIndex` value of the last sub-processor in the list, even though we are drawing any other processor and not the last one. I tested your PR with a neuropixels module with 2 probes attached to it, and found out that this causes performance issues and CPU usage spikes every time an event-triggered display occurs. It also makes the GUI freeze sometimes, especially when the sub-processors you are switching between have different sample rates. I have recorded a video of my test, you can watch that video via this link : https://drive.google.com/file/d/1KjJPVZrE8Z5rNMPrxUTqiBx818zj3PFv/view

That is why we only need to set the `lastTrigger` value for the sub-processor we are currently drawing. We need to move the that segment of code outside the for loop. So, my suggested change looks something like this:

```suggestion
    }
    if (latestCurrentTrigger >= 0)
    {
        int chan = numChannelsInSubprocessor[subprocessorToDraw];
        int subProcIndex = allSubprocessors.indexOf(subprocessorToDraw);
        latestTrigger = latestCurrentTrigger + displayBufferIndices[subProcIndex][chan];
    }
```
With this change, I'm able to get rid of the performance issue I was facing earlier, and everything works fine. My question is, should I merge the PR as-is and then make my changes right after, or do you want to test this change first and then add a new commit to this PR?

Also, in LfpDisplayCanvas.cpp, you say that "trigger event is positioned at 1/3 of the display width", but that is not the case. In reality, the trigger event is positioned at 1/4 of display width. Can you please confirm this?

wagenadl:
Hi Anjal,

Thanks for working on the pull request. I had not tested my code in the
context of multiple processors, and I am glad that you did. Currently, I
am away from the lab and am therefore not in a position to do any
testing with live data. As a consequence, I am happy for you to make the
changes directly. (Also, I am not sure I quite understand the
intricacies of the multi-processor arrangement.)

You are right, the trigger is at 1/4. Sorry for the incorrect
documentation. If you think 1/3 is better, it is of course easy to change.

Best, Daniel

anjaldoshi:
Okay, no worries. I'll merge this PR and then add the changes mentioned above after that. 

I think trigger at 1/4 looks fine for now, but I'll consult with my team and see what they say.

Thanks for your continued contribution!

PR #364, opened 2020-08-26
Convert Info Tab into a webview component

jsiegle:
Screen Shot 2020-08-26 at 10 51 21 AM

Summary of changes:
- Make the InfoLabel into a WebBrowserComponent instead of a standard Juce component, so we can render HTML pages
- Generate HTML + CSS to display "About," "Authors," and "License"
- Update CMakeLists.txt to copy the relevant files into an "assets" directory

Still needed:
- testing on Windows and Linux


jsiegle:
Looks like this is going to be a pain to implement on Linux and Windows without upgrading to Juce 6.

PR #367, opened 2020-09-02
Lift requirement that sinks terminate a signal chain

jsiegle:
These updates make it so that sinks can be placed in the middle of a signal chain, similar to how filters currently behave. 

Other changes include:
- Simplified logic for drawing GraphView
- New aesthetics for GraphView (see below)
- Fixed bugs when moving around Splitters or Mergers
- Add a distance threshold for dragging editors, to avoid accidentally triggering a "MOVE" event

Screen Shot 2020-09-02 at 2 35 44 PM

### DIFF
--- a/Source/Processors/GenericProcessor/GenericProcessor.cpp
+++ b/Source/Processors/GenericProcessor/GenericProcessor.cpp
@@ -176,30 +176,30 @@ void GenericProcessor::setSourceNode(GenericProcessor* sn)
 		{
 			//	std::cout << " The source is not blank." << std::endl;
 
-			if (!sn->isSink())
-			{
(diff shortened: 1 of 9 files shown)


jsiegle:
Just pushed some changes that improve Merger functionality:
- Now, a Merger can only be connected to processors at the end of a signal chain, and which are not downstream from itself. This prevents the creation of weird loops, which were previously possible
- Right-clicking on a Merger brings up a pop-up menu that now includes info about _both_ of its inputs. If an input is already connected, the user can toggle whether to send events or continuous data. If there is no input connected, the user can select which ones to merge (if there are mergeable processors available):

Screen Shot 2020-09-04 at 10 38 27 AM

PR #393, opened 2020-11-30
Prevent using OpenEphys format w/ > 300 recorded channels

medengineer:


jsiegle:
You should add another check for this in `RecordNode::enable()`, because it's possible to add a high-channel-count source after the Open Ephys format has already been selected.

Also changed "OpenEphys" > "Open Ephys"

jsiegle:
Also, shouldn't it be `box->setSelectedItemIndex(1);` for the Binary format?

Issue #410, opened 2021-01-29
Сrash on startup


Alkohost:
I just upgraded to the latest version but still face with crash when launch the GUI (win7). 
It happens if 
1) the configuration is changed (if one of cables is disconnected) but option Reload on startup is enabled (in the previous version I just manually deleted previous config before start the GUI). 
2) the board power is off and I press cancel.

Is it possible to fix this crash problems with coming software updates?

jsiegle:
Thanks for reporting this! We will definitely get these fixed.

Alkohost:
Thank you. As far as I remember, there was also a crash due to bug/feature if a PC did not have an audio output connected. Is it fixed already?

Сan I ask questions about hardware here or should I write to the /acquisition-board ?

jsiegle:
This is fixed in the latest release (0.5.3.1).

Regarding the audio output, the GUI needs to be able access an audio output device, otherwise it won't be able to function (all of the data acquisition callbacks are driven by the audio card).

PR #411, opened 2021-01-30
Prevent crashes when re-loading signal chain with disconnected acq board

jsiegle:
Cases tested:
- 2 headstages + ADCs in saved state, 1 headstage on re-load
- 1 headstage + ADCs in saved state, no headstages on re-load
- ADCs in saved state, no acquisition board connected on re-load

Also made some changes to prevent LFP Viewer from crashing with 0 input channels.


jsiegle:
Closes #410

PR #419, opened 2021-03-01
Add HARP bus capabiltiies to bitfiles

aacuevas:
Add the ability to use the acquisition board as a master of the HARP synchronization bus, developed by the Champalimaud fundation. Uses the now unused i2c data/clk lines in the IO boards.


jsiegle:
Very cool! We will get this merged before the next release.

PR #427, opened 2021-04-05
Fix sync_messages.txt intermittently missing entries

medengineer:


jsiegle:
What's the difference between this and #414?

medengineer:
No difference, accidental duplicate of #414. Closing.

PR #474, opened 2021-08-30
Fix inconsistencies between UI and processor in Bandpass Filter

stinos:


stinos:
I noticed other places use `auto` so rewrote some parts to use C++11 features.

PR #476, opened 2021-08-31
Fix nullptr dereferencing when plugin not found

stinos:
Don't call setEnabledState() in the editor's constructor because it
forwards to the processor's setEnabledState() which in turn calls the
editor's enable(), but editor is still nullptr at that point since
it's being constructed.
Doesn't change the initial intent of creating the plugin in a disabled
state but fixes crash when loading configurations which refer to
non-existing plugins for instance.


anjaldoshi:
Hi @stinos,

Thank your for the contributions. They all look useful. We'll merge these into the `development-juce6` branch before the next release (0.6.0), expected in October. 

\- Anjal

stinos:
@anjaldoshi ok thanks, but please review carefully: codebase is new to me so there might be better/more correct ways to do things.

anjaldoshi:
Yep, will surely do! Since v0.6.0 has some major updates, including updating the JUCE API to version 6 (we are currently using v4.2.1), we'll have to do some extensive reviews and testing and we'll keep this in mind.

PR #478, opened 2021-09-21
Fixing some bugs responsible for seemingly random crashes

jsiegle:
Fixes a crash that's happening whenever a key press event is sent to the Spike Viewer. Pressing "c" no longer clears the active spikes, but this can also be accomplished by pressing the "clear plots" button.


jsiegle:
Seems like this is only causing crashes when running in debug mode, due to a leaked KeyPress object

jsiegle:
Adding some more fixes:
- The `while` loop in `ProcessorGraph::updateViews()` was randomly getting stuck – changing this to a `for` loop prevents crashes
- `WaveAxes::plotSpike()` can get into a weird state while acquisition is not active, or when just starting acquisition. Added an `if` statement that checks for outlandish values before plotting
- When connecting processors before starting acquisition, it's possible for the Splitter's internal state to changes. Added a fix that restores each Splitter's original state.
- When a processor is dragged outside the EditorViewport, it's now deleted instead of moved. This prevents crashes when the insertion point is ambiguous.

PR #514, opened 2022-06-24
Update HTTP directory field

bjhardcastle:
`base_text` looks like the correct var name

`current_directory_name` appears unused

[Fix directory naming via HTTPServer · open-ephys/plugin-GUI@626f50d · GitHub](https://github.com/open-ephys/plugin-GUI/commit/626f50dc8a082808a56783e1ae5ea47ee0e4dcf8#diff-cf1044d55166de06cac7f30258205bd55a4f80593c69c3809fa458c142c2c07dR193)

### DIFF
--- a/Source/CoreServices.cpp
+++ b/Source/CoreServices.cpp
@@ -142,9 +142,14 @@ namespace CoreServices
 		return getControlPanel()->getRecordingParentDirectory();
 	}
 
-	void setRecordingDirectoryBasename(String dir)
+	void setRecordingDirectoryBaseText(String text)
 	{
-		getControlPanel()->setRecordingDirectoryBasename(dir);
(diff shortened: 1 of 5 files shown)


jsiegle:
Hi @bjhardcastle ...thanks for submitting this fix. I think we actually need to add a new `CoreServices` method called `getRecordingDirectoryBaseText()`, and have this be what gets returned in the `base_text` field. That way you can validate whether the requested changes to the base text took effect, while also seeing the current recording directory name (which may be auto-generated). Would that solve the issue?

bjhardcastle:
Hi Josh/Pavel,
This looks perfect - thank you for implementing the changes already. 
I might attempt to rebuild and start using it this week, but setting the name in "open-loop" without validation is working fine for now. 
Thanks!

PR #518, opened 2022-08-10
V0554 intan fixes

cjthomas-opensource:
Minor bug fixes without changes to functionality (cable length fixes will be a separate pull request).
- Fixed channel counts not rendering for ports E..H.
- Fixed initialization of placeholder cable lengths for ports E..H (they were from lengths A..D).
- Fixed constructor initialization of cable length variables (lengths E..H weren't being initialized).
- Moved default cable length magic value to a header macro.
- Moved number of GUI port slots magic value to a header macro (and fixed inconsistent values per rendering bug).
- Replaced sampling rate index magic value with enum value per other sampling rate indices.
Bench-tested with a recording controller and headstage, but not stress-tested.

### DIFF
--- a/Plugins/IntanRecordingController/RHD2000Editor.cpp
+++ b/Plugins/IntanRecordingController/RHD2000Editor.cpp
@@ -590,7 +590,7 @@ RHD2000Editor::RHD2000Editor(GenericProcessor* parentNode,
 	impedanceData->valid = false;
 
     // add headstage-specific controls (currently just an enable/disable button)
-    for (int i = 0; i < 8; i++)
+    for (int i = 0; i < INTAN_EDITOR_PORT_SLOTS; i++)
     {
         HeadstageOptionsInterface* hsOptions = new HeadstageOptionsInterface(board, this, i);
         headstageOptionsInterfaces.add(hsOptions);
@@ -834,7 +834,7 @@ void RHD2000Editor::buttonEvent(Button* button)
     {
         board->scanPorts();
 
-        for (int i = 0; i < 4; i++)
+        for (int i = 0; i < INTAN_EDITOR_PORT_SLOTS; i++)
         {
             headstageOptionsInterfaces[i]->checkEnabledState();
         }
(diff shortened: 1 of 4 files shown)


cjthomas-opensource:
I've modified cable delay handling significantly in the most recent commit. Noteworthy changes:
- The port scan is performed at the user-configured sampling rate (unless overridden, which happens during initialization). This avoids having to save cable lengths and do a delay to length to delay double conversion. Cable lengths are still saved but not used.
- A port scan is performed whenever the user changes the sampling rate, or toggles a delay adjustment button. This takes negligible time and ensures that cable delays are always properly configured.
- There's a DELAY_DEBUG switch to give more succinct output than the SCAN_DEBUG switch. A screenshot of the output is attached.
- The Intan GUI has a button for each port to toggle delay adjustment (screenshot attached). Delay adjustment is either "0" or "+1", toggled by pressing the button. In my tests these were the only adjustments needed (software defaults to the lowest delay when exactly two delays are detected, and this was the main case that showed up).
- I added a HS_PANEL_WIDTH macro to the editor equal to twice HS_WIDTH. Way too many magic constants were assuming a single column and breaking when HS_WIDTH was changed. These have been updated.
delay-debug-output
intan-recorder-with-delay-adjust

cjthomas-opensource:
Here text files describing my test procedure, and screenshots of what clean and corrupted signals look like. Digital corruption occurs when the cable delay isn't set correctly; bit flips happen in the data packets giving large stepwise changes in the data.

[20220809-test-procedure.txt](https://github.com/open-ephys/plugin-GUI/files/9312415/20220809-test-procedure.txt)
[20220811-test-results.txt](https://github.com/open-ephys/plugin-GUI/files/9312416/20220811-test-results.txt)
signal-with-bad-connection
signal-with-bad-connection-low-amplitude
signal-with-good-connection

jsiegle:
Thanks for making these changes! We aren't planning to put out a new release for GUI `v0.5.x`, but we will make sure these changes get incorporated into the next version of the [Rhythm Plugins](https://github.com/open-ephys-plugins/rhythm-plugins), so they can be used with the Open Ephys Acquisition Board as well as the Intan Recording Controller.

PR #527, opened 2022-09-28
[OpenEphysHttpServer] Add support for adding and deleting a processor

anjaldoshi:
Add support for adding and removing a processor from the signal chain via the OpenEphysHttpServer.

New PUT requests: 

| URL | /api/processors/add |
| --- | --- |
| json | "**name**": "_processor_display_name_",  "**source_id**": "_source_processor_id_" OR "**dest_id**": "_destination_processor_id_" |
| comments | "name" should match the plugin display name. Either source or destination processor node id is required. |

| URL | /api/processors/delete |
| --- | --- |
| json | {"**id**": "_processor_node_id_"} |
| comments | "id" should be the processor node id. |

Also, added a helper function to get plugin description for an item in the Processor List.

### DIFF
--- a/Source/UI/ProcessorList.cpp
+++ b/Source/UI/ProcessorList.cpp
@@ -716,6 +716,36 @@ Array<String> ProcessorList::getItemList()
 	return listOfProcessors;
 }
 
+
+Plugin::Description ProcessorList::getItemDescriptionfromList(const String& name)
+{
+
+	Plugin::Description description;
+    
+	for (int i = 0; i < 5; i++)
+	{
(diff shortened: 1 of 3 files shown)


jsiegle:
I had to change a few things to get this to work:

- Add `MessageManagerLock` just before adding or removing processors
- Parse processor IDs as integers, rather than strings
- Handle case in which source or dest node ID is invalid
- Handle case in which there's a disagreement between source/dest node IDs (e.g., only rely on one or the other for processor placement, but not both)

Please test the updated version and make sure everything looks ok!

jsiegle (approved):

PR #550, opened 2023-02-06
V0554 arduino dev

cjthomas-opensource:
Quality-of-life improvements to v0.5 Arduino Output plugin. It properly handles TTL inputs in multiple streams, gives human-readable output channel names, loads/saves configuration properly, and caches configured signal selections by human-readable name to attempt to rebuild configuration after geometry changes (so that you don't necessarily lose configuration when editing signal chain components upstream).

arduino-output-mod-20230201


medengineer:
Thanks for the update! Just added a small commit on top that fixes some spacing for messages sent via CoreServices.

PR #551, opened 2023-02-07
Adding lfp channel selection filtering

admunkucsd:
-Added filteredChannels StringArray to DisplayBuffer that holds a list of channels to show from LFP display
-Added broadcastMessage parsing that expects JSON formatted message to describe a rectangular selection of channels to filter

### DIFF
--- a/Plugins/LfpDisplayNode/LfpDisplayNode.cpp
+++ b/Plugins/LfpDisplayNode/LfpDisplayNode.cpp
@@ -180,6 +180,11 @@ bool LfpDisplayNode::stopAcquisition()
     LfpDisplayEditor* editor = (LfpDisplayEditor*) getEditor();
     editor->disable();
 
+    for(auto split : splitDisplays) {
+        Array<int> emptyArray = Array<int>();
+        split -> setFilteredChannels(emptyArray);
+    }
+    
     for (auto buffer : displayBuffers)
         buffer->ttlState = 0;
 
--- a/Plugins/LfpDisplayNode/LfpDisplayNode.cpp
+++ b/Plugins/LfpDisplayNode/LfpDisplayNode.cpp
@@ -327,3 +327,20 @@ void LfpDisplayNode::acknowledgeTrigger(int id)
 {
     latestTrigger.set(id, -1);
 }
+
+String LfpDisplayNode::handleConfigMessage(String msg) {
+    std::cout << "in lfp config: " << msg <<std::endl;
+}
+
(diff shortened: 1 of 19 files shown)


jsiegle:
Thanks for this! I tested this out using a Python script, formatting the message as a dictionary and using `json.dumps` before sending, e.g.:

```python

from open_ephys.control import OpenEphysHTTPServer
import json

gui = OpenEphysHTTPServer()

message = {
    'plugin' : 'LFPViewer',
    'command' : 'filter',
    'payload' : 
    {
        'streamID' : 10001,
        'start' : 2,
        'rows' : 3,
        'cols' : 1,
        'colsPerRow' : 0
    }
}

gui.message(json.dumps(message))
```

This approach worked right out of the box, which was great!

I have a few suggestions for making the filtering process more general:
- I think we should avoid using `streamID` here, since that is only accessible internally, and can change when the GUI is reloaded. Instead, the payload should specify the `splitID` to update (0, 1, or 2). That way you could have multiple active views of the same stream, but only filter one of them. Also, when I specify an invalid `streamID`, it currently crashes.
- Related to the last point, the list of channels to filter should be updated directly in the split displays. There's no reason to go through the DisplayBuffer for this.
- It would be helpful to have an easy way to filter channels that doesn't involve rows and columns. For example, if an `end` parameter is set, it just selects channels between `start` and `end` and ignores `rows` and `cols`.
- It's crashing a lot when test out different payload parameters -- it needs to be robust to all possible inputs.

Plugins/LfpDisplayNode/LfpDisplayNode.cpp
```diff
@@ -329,6 +329,17 @@ void LfpDisplayNode::acknowledgeTrigger(int id)
     latestTrigger.set(id, -1);
 }
 
+bool LfpDisplayNode::getIntField(DynamicObject::Ptr payload, String name, int& value, int lowerBound, int upperBound) {
```
admunkucsd, line 332: I added this to Lfp for now but I think it would be nice to templatize this move this (and some other tools) into a Message Utility class when we can update the Plugin API.

jsiegle: I incorporated this method into the Online PSTH plugin, and had to change one line to get it to compile on Linux: https://github.com/open-ephys-plugins/online-psth/commit/789ad31615b3dd814a28bd2def1bfbfd02a5ae27

PR #554, opened 2023-02-10
V0554 recordnode dev

cjthomas-opensource:
Modified RecordNode to have more consistent buffer sizes. This greatly improved stability when trying to record on an I/O-bottlenecked system and with small audio buffer sizes (needed for closed-loop work).

Modified BinaryFormat to use fewer buffers that are much larger. This greatly improved performance when writing to magnetic platter drives. RAM footprint is still acceptable (2 MB/channel per buffer block, 16 MB/channel total pre-allocated).

These changes have been tested with a single 128ch headstage running for several minutes, saving two copies of the data to a magnetic platter disk array at the same time, with an audio buffer duration of 6 ms. A full description of all of the test cases (with and without the code changes) is available upon request.

Long-duration tests (from actual research recording sessions) should be in-hand in the near future, but the bench tests should be sufficient for showing that no bugs were introduced.

### DIFF
--- a/Source/Processors/RecordNode/BinaryFormat/BinaryRecording.cpp
+++ b/Source/Processors/RecordNode/BinaryFormat/BinaryRecording.cpp
@@ -41,8 +41,6 @@ int BinaryRecording::getSamplesPerBlock()
 	case BINBLOCK_LARGE: blockSamps *= 16; break;
 	default: break;
 	}
-// FIXME - Diagnostics.
-std::cerr << "[BinaryRecording]  getSamplesPerBlock() returns: " << blockSamps << "\n";
 
 	return blockSamps;
 }
--- a/Source/Processors/RecordNode/BinaryFormat/BinaryRecording.cpp
+++ b/Source/Processors/RecordNode/BinaryFormat/BinaryRecording.cpp
@@ -4,6 +4,8 @@
 
 #define MAX_BUFFER_SIZE 40960
 
+#define MIN_SAMPLES_PER_BLOCK 65536
+
 BinaryRecording::BinaryRecording()
 {
     m_bufferSize = MAX_BUFFER_SIZE;
@@ -26,6 +28,23 @@ String BinaryRecording::getProcessorString(const InfoObjectCommon* channelInfo)
     fName += "." + String(channelInfo->getSubProcessorIdx());
 	fName += File::separatorString;
 	return fName;
+}
+
+int BinaryRecording::getSamplesPerBlock()
+{
+	int blockSamps = MIN_SAMPLES_PER_BLOCK;
+
+	switch (m_blockSizeBoost)
+	{
+	case BINBLOCK_SMALL: break;
+	case BINBLOCK_MEDIUM: blockSamps *= 4; break;
+	case BINBLOCK_LARGE: blockSamps *= 16; break;
+	default: break;
+	}
+// FIXME - Diagnostics.
+std::cerr << "[BinaryRecording]  getSamplesPerBlock() returns: " << blockSamps << "\n";
+
+	return blockSamps;
 }
 
 void BinaryRecording::openFiles(File rootFolder, int experimentNumber, int recordingNumber)
@@ -120,6 +139,8 @@ void BinaryRecording::openFiles(File rootFolder, int experimentNumber, int recor
         }
         lastId = indexedDataChannels.size();
     }
+
+    int samplesPerBlock = getSamplesPerBlock();
 
     int nFiles = continuousFileNames.size();
     for (int i = 0; i < nFiles; i++)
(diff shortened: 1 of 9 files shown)


jsiegle:
Thanks for this PR! 

I folded these changes into the latest `development` branch and tested recording with 8 simulated Neuropixels probes (3072 channels at 30 kHz and 3072 channels at 2.5 kHz) writing to an SSD with a 21 ms sample buffer. This signal chain worked fine previously, but with the new recording parameters, it started lagging after about 30 seconds (CPU meter at 100%, buffer starting to fill). I saw the same thing for 4 simulated probes writing to two consecutive Record Nodes, with a 6 ms sample buffer.

Before we decide whether to incorporate these changes, it would be helpful to have a better understanding of how the data queue and block size parameters affect recording performance under more typical use cases (SSD, larger sample buffers).

cjthomas-opensource:
My best guess is that this is due to it running out of memory and hitting swap.

There are two places where I changed the size of allocated buffers. In RecordNode.h/RecordNode.cpp, I changed the number of blocks requested when creating DataQueue objects so that the queue always had at least 100 blocks and at least 1M samples per channel (old behavior was fixed at 300 blocks). With a 21 ms audio buffer, channels sampled at 30 ksps had blocks 640 samples long, and channels sampled at 2.5 ksps had blocks about 53 samples long. The old behavior would allocate 192k samples per 30 ksps channel (300 x 640) and about 16k samples per 2.5 ksps channel, for a total of about 640M samples, or about 2.6 GB when stored as single-precision floating point. The new behavior would allocate a flat 1M samples per channel for all 6144 channels, or about 25 GB when stored as single-precision floating-point.

The other place I changed the size of buffers was in BinaryRecording.h. The old behavior used 128 blocks of 4k samples (with samples stored as 16-bit integers), and the new behavior uses 8 blocks of 1M samples. The old behavior's footprint would pre-allocate 0.5M samples (1 MB) per channel, for a total of 6 GB for the write buffers. The new behavior would pre-allocate 8M samples (16 MB) per channel, for a total of 98 GB for the write buffers.

So, with 6k channels, it's asking for about 123 GB of memory. I recommend using a computer with at least 192 GB of RAM (and preferably 256 GB) if you're doing that with my version of the record node. I'm guessing your machine has less than that, which means it'll allocate the buffers in swap, behave normally as long as its working set is smaller than physical RAM, and then start thrashing as soon as its actual buffer use becomes larger than that.

In theory, it should only ever actually keep two out of eight write buffers in RAM, since it'll flush the previous one as soon as it starts a new one (and the write buffer size is much larger than the DataQueue size so it shouldn't overlap multiple buffers at once), but in practice it isn't actually doing that: I'd tested with 4 write buffers instead of 8, but found that it misbehaved.

The buffer geometries that I chose were specifically geared towards my test cases (needing to do closed-loop work with very short audio buffer duration and about 1k channels). I recognize that settings that are optimal for that may be less optimal for multiple thousands of channels. While I'd assumed that anyone speccing a machine to acquire from four neuropixels at once would stuff in as much RAM as the architecture supported, I realize that you could say the same thing about stuffing it full of SSDs.

How would you feel about me adding a selector so that the user can choose the RecordNode buffer sizes, with the default configuration having a footprint closer to the original version? This should let your setup work out of the box while still letting me run my setup without custom code (assuming no other surprises turn up during testing).

jsiegle:
I think adding a selector should solve the problem for now. Eventually we'll want to do more benchmarking to see if these parameters can be optimized further (and perhaps set the queue/block sizes dynamically based on the number of simultaneously recorded channels).

cjthomas-opensource:
I've added a combobox to select buffer size; a picture is attached. The "small" buffer has the same memory footprint as the stock setup; it just uses 8 write buffers of size 64k instead of 128 write buffers of size 4k. The "large" buffer has the same footprint as my original modification. There's also an intermediate "medium" size.

This has been bench-tested using my previous stress tests, and has been stability-tested by recording from a 768-channel Intan-based setup. All of these tests were done with magnetic platter drives.

record-node-20230214

Source/Processors/RecordNode/BinaryFormat/BinaryRecording.cpp
```diff
@@ -28,6 +30,23 @@ String BinaryRecording::getProcessorString(const InfoObjectCommon* channelInfo)
...
+
+	switch (m_blockSizeBoost)
+	{
+	case BINBLOCK_SMALL: break;
+	case BINBLOCK_MEDIUM: blockSamps *= 4; break;
+	case BINBLOCK_LARGE: blockSamps *= 16; break;
+	default: break;
+	}
+// FIXME - Diagnostics.
+std::cerr << "[BinaryRecording]  getSamplesPerBlock() returns: " << blockSamps << "\n";
```
jsiegle, line 45: Change this to `LOGD`

cjthomas-opensource: I've removed it instead (I'd missed it when removing other debugging tattles).

PR #578, opened 2023-07-07
Add more unit tests; have test infra use ProcessBlock for Plugin tests

admunkucsd:
Contributions by @pbotros and @admunkucsd 
- Change Unit Tests to instance a `ProcessorTester` object that manages `Components` and `Processors` outside of the UUT
- Add Unit Tests for:
  - `DataViewport`
  - `ProcessorGraph`
  - `RecordNode`
  - `ChannelMap`
  - `LFPViewer`
  - `FilterNode`


Source/Processors/SourceNode/SourceNode.h
```diff
@@ -61,6 +61,9 @@ class PLUGIN_API SourceNode : public GenericProcessor
 
     /* Broadcasts a message from the DataThread to all other processors*/
     void broadcastDataThreadMessage(String msg);
+    
+    /* Broadcasts a message from the DataThread to all other processors*/
+    void sendDataThreadConfigMessage(GenericProcessor* destProcessor, String msg);
```
pbotros, line 66: why have this alias of `GenericProcessor::sendConfigMessage`? Seems more straightforward to have callers call `sendConfigMessage` directly

Source/Processors/Visualization/Visualizer.cpp
```diff
@@ -40,7 +40,7 @@ void Visualizer::update()
 
     if (processor != nullptr)
     {
-        LOGDD(processor->getEditor()->getNameAndId(), " visualizer updating parameter editors");
+//LOGDD(processor->getEditor()->getNameAndId(), " visualizer updating parameter editors");
```
pbotros, line 43: Intentional?

Source/UI/ControlPanel.cpp
```diff
@@ -52,7 +52,7 @@ PlayButton::PlayButton()
     setClickingTogglesState(true);
     setTooltip("Start/stop acquisition");
     
-    updateImages();
+    //updateImages();
```
pbotros, line 55: Intentional?

Source/UI/ControlPanel.cpp
```diff
@@ -89,7 +89,7 @@ RecordButton::RecordButton()
     setClickingTogglesState(true);
     setTooltip("Start/stop writing to disk");
     
-    updateImages();
+    //updateImages();
```
pbotros, line 92: Intentional?

Source/Utils/BroadcastParser.cpp
```diff
@@ -0,0 +1,73 @@
...
+    int tempVal = payload->getProperty(name);
+    if((upperBound != INT_MIN && tempVal > upperBound) || (lowerBound != INT_MAX && tempVal < lowerBound))
+        return false;
+    value = tempVal;
+    return true;
+}
+
+
+
+bool BroadcastParser::checkForCommand(String expectedPlugin, String expectedCommand, String msg, var& payload) {
```
pbotros, line 39: this functionality looks more like this function should be called `getPayloadForCommand()`. Also document in the header what the `bool` return value means for this function and other functions

Source/Utils/BroadcastParser.cpp
```diff
@@ -0,0 +1,73 @@
...
+    }
+    var tempPayload = jsonMessage -> getProperty("payload");
+    if(tempPayload.getDynamicObject() == nullptr){
+        return false;
+    }
+    payload = jsonMessage -> getProperty("payload");
+    return true;
+}
+
+GenericProcessor* BroadcastParser::getDestinationNode(GenericProcessor * root, int nodeId) {
```
pbotros, line 64: this function doesn't look related at all to the BroadcastParser. Does this exist somewhere else or should it be moved somewhere else, perhaps towards ProcessorGraph or AccessClass or something?

Source/Utils/BroadcastParser.cpp
```diff
@@ -0,0 +1,73 @@
...
+    jsonObj->setProperty("command", command);
+    jsonObj->setProperty("payload", var(payloadObj));
+    String returnString = JSON::toString(var(jsonObj), true);
+    payloadObj.reset();
+    jsonObj.reset();
+    return returnString;
+}
+
+
+bool BroadcastParser::getIntField(DynamicObject::Ptr payload, String name, int& value, int lowerBound, int upperBound) {
```
pbotros, line 27: Inevitably it seems like there's going to be many different types of fields besides `int`, and interpreting the payload has more to do with the payload itself than the broadcast parser.

A more straightforward way to structure this class for developers would be to return a class encapsulating the payload, perhaps named `BroadcastPayload`, that has member methods like `getIntField(name, value, ...)`. You could then never return the raw `payload` (i.e. like in `checkForCommand`) and instead always return this wrapped payload class. This also doubles as giving us flexibility in the future to change the underlying data structure of the broadcasted payloads, as JSON is fairly slow and not amenable to large binary payloads.

TestHelpers/include/TestFixtures.h
```diff
@@ -2,40 +2,232 @@
...
+        control_panel->updateRecordEngineList();
+
+        // Refresh everything
+        processor_graph->updateSettings(sn, false);
+    	
+        control_panel ->updateColors();
+
+	}
+
+    ProcessorTester(DataThreadCreator data_thread_creator) {
```
pbotros, line 54: The two constructors are very similar. Can you consolidate the two so that there's less copy/paste and thus less room for the implementations to diverge?

TestHelpers/include/TestFixtures.h
```diff
@@ -2,40 +2,232 @@
...
+    std::unique_ptr<ProcessorGraph> processor_graph;
+    std::unique_ptr<ControlPanel> control_panel;
+
+    int next_processor_id_ = 1;
+    int current_sample_index = 0;
+    std::unique_ptr<CustomLookAndFeel> customLookAndFeel;
+};
+
+#define _USE_MATH_DEFINES
+#include <math.h>
```
pbotros, line 221: put includes up top

TestHelpers/include/TestFixtures.h
```diff
@@ -2,40 +2,232 @@
...
+
+class ProcessorTest : public ::testing::Test {
```
pbotros, line 223: Is this used anymore? If not, let's remove it.

Tests/Processors/ProcessorGraphTests.cpp
```diff
@@ -0,0 +1,146 @@
...
+private:
+    std::unique_ptr<AudioComponent> audio_component;
+    std::unique_ptr<ControlPanel> control_panel;
+};
+
+TEST_F(ProcessorGraphTest, LoadFromXMLTest) {
+    
+    std::string default_file_reader_path = "default";
+    File currentPath = File::getSpecialLocation(File::currentApplicationFile);
+    while(currentPath.getFileName() != "plugin-GUI") {
```
pbotros, line 41: this is a bit fragile - what if people clone their repo as something else? Any other way to bake this in, perhaps using a variable we can inherit from CMake?

Alternatively, if not, this should have some exit condition so it doesn't loop forever in the case of bugs.

Tests/Processors/ProcessorGraphTests.cpp
```diff
@@ -0,0 +1,146 @@
...
+        currentPath = currentPath.getParentDirectory();
+    }
+    
+    currentPath = currentPath.getChildFile("Resources/FileReader/resources/structure.oebin");
```
pbotros, line 45: would this work on Windows? Since this uses forward slashes instead of backslashes.

Tests/UI/DataViewportTests.cpp
```diff
@@ -0,0 +1,62 @@
+#include "gtest/gtest.h"
+#include "../../Source/UI/DataViewport.h"
+#include "../../Source/UI/InfoLabel.h"
+
+class DataViewportTests : public ::testing::Test {
+protected:
+    void SetUp() override {
+//        MessageManager::deleteInstance();
```
pbotros, line 8: remove commented code

Source/Processors/SourceNode/SourceNode.h
admunkucsd: I made this wrapper for cases when `DataThread` plugins needed to send a config message, as the Thread's `SourceNode` is what actually derives from `GenericProcessor`. `sendBroadcastMessage` and `sendConfigMessage` are both protected members of `GenericProcessor`, but, after looking at this again, these `SourceNode` wrappers do defeat the purpose of keeping these members proctected.

I believe `MessageCenter` has a lot `Thread` safety checks, so careless use of message passing on non-`ProcessorGraph` threads can lead to unintended consequences. If this we want to enforce this access limitation, which seems reasonable given the intention of cross-processor communication, then perhaps `DataThread` should be a friend of `SourceNode`? Or maybe there is a better way to  define this relationship?

admunkucsd:
Addressed feedback in [2c5b262](https://github.com/open-ephys/plugin-GUI/commit/2c5b262c9a3d9c6dd1da13de85c1bdbc9fe15727)

pbotros (approved):
LGTM!

PR #613, opened 2024-06-19
merge ug3 core application changes

benjaminmao123:
- LFPViewer
- > Added Tooltip for channels so users can see channel names that are cut off.
- > Increased margin to 100
- > Fixed handleBroadcastMessage logic to support subselect for UG3 Electrode Viewer plugin.

- CoreServices
- > Added methods to retrieve the processor by name/id
- > Added method to retrieve predecessor processor ids

- DataBuffer
- > Added an additional parameter to DataBuffer to store the timestampSampleIndex which is used by our Synchronizer plugin.

- Event
- > Added an additional fillReferenceSampleEvent which is used by our Synchronizer plugin.

- GenericProcessor
- > Added a method to set the reference sample for a block which is used by our Sychronizer plugin.

- RecordNode
- > Added a method to override the record engine for testing

- BroadcastPayload
- > Added a default constructor
- > Added a getter for the payload.


jsiegle:
It looks like there are going to be a ton of conflicts here. Instead of directly merging these two branches, can you copy and paste the relevant changes onto the latest `development-juce8` branch?

Issue #615, opened 2024-06-20
Incorrect std value in LFP viewer 


florian6973:
Hi!

I generated a white noise signal with a std of 10. When playing it in OE, I get absurd std values like 1-2 instead of something around 10 (see image below).
![image](https://github.com/open-ephys/plugin-GUI/assets/70778912/2d7175e5-e630-4578-916e-64760f5a263c)

Am I missing something please?

I thank you in advance for your help,

Best,

florian6973:
Here is one of the files if you'd like to test
[structure5.zip](https://github.com/user-attachments/files/15932378/structure5.zip)

jsiegle:
Thanks, I see what the problem is – the [standard deviation is being calculated on the "screen buffer"](https://github.com/open-ephys/plugin-GUI/blob/8c6453954b431a1f48f18b8fc3ddd1282eeda56d/Plugins/LfpDisplayNode/LfpDisplayCanvas.cpp#L1569-L1584) (which aggregates multiple samples into a single pixel), rather the "display buffer" (which contains the original samples). We are planning to overhaul this part of the LFP Viewer in the next release (and likely switch to displaying RMS values, instead of standard deviation). We will make sure this gets fixed!

florian6973:
Thank you very much for your insights! Computing RMS or MAD values on several channels at the same time would be a very exciting feature indeed!

In the meantime, is there an easy way to replace the screen buffer by the display buffer in the code (like rewriting getMean and getStd with `displayBuffer` instead of `screenBufferMean`) and recompile or is it better to use a custom plugin or custom script with ZeroMQ streaming to measure noise please?

jsiegle:
I think actually just changing those functions to read from the `displayBuffer` should work, as long as you get the indexing right. It should read around 0.1 s worth of samples up to the current `displayBufferIndex` for the selected channel. Can you try that and see if it gives reasonable values?

florian6973:
Thanks for your help!

I updated the code accordingly https://github.com/florian6973/plugin-GUI/commit/e5d92ba4009cb42b76f974c98f608bf1e66a987f. It seems to work almost well:
- the std value now makes sense (10 expected):
![image](https://github.com/open-ephys/plugin-GUI/assets/70778912/823782dd-bea1-4b33-95a1-a97661974b43)
- but sometimes when refreshing by a click it just displays 0, I have to click again to get a new coherent value:
[...]

PR #616, opened 2024-06-26
LFPViewer - fix mean+std calculation

florian6973:
Regarding this issue https://github.com/open-ephys/plugin-GUI/issues/615, here is a small fix of the mean and std calculation, before a bigger refactoring of the LFP viewer.

### DIFF
--- a/Plugins/LfpDisplayNode/LfpChannelDisplay.cpp
+++ b/Plugins/LfpDisplayNode/LfpChannelDisplay.cpp
@@ -243,7 +243,7 @@ void LfpChannelDisplay::pxPaint()
         double a = (canvasSplit->getYCoordMax(chan, index)/range*channelHeightFloat);
         double b = (canvasSplit->getYCoordMin(chan, index)/range*channelHeightFloat);
             
-        double mean = (canvasSplit->getMean(chan)/range*channelHeightFloat);
+        double mean = (canvasSplit->getScreenBufferMean(chan)/range*channelHeightFloat);
             
         if (drawWithOffsetCorrection)
         {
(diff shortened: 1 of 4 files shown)


anjaldoshi:
Hello @florian6973,

We appreciate your contribution. Upon reviewing your changes, I found that the Mean and Std Dev values are now accurately calculated using the data from the display buffer instead of the screen buffer. However, we noticed an issue when the “subtract offset” option is enabled. The signal plotted seems a bit off. This is due to the `LfpDisplaySplitter::getMean` function being used not only to calculate and display the channel-specific mean and std values, but also to adjust the signal’s Y coordinate’s min and max values for offset correction (you can refer to the code [here](https://github.com/open-ephys/plugin-GUI/blob/development/Plugins/LfpDisplayNode/LfpChannelDisplay.cpp#L242-L252)). As the offset correction is performed when plotting the signal using the screen buffer, the mean value used there is incorrect because it’s derived from the display buffer with your modifications.

To rectify this, I suggest updating your current PR. The optimal solution would be to have two distinct methods for calculating the mean: one for drawing/plotting that utilizes the screen buffer, and another for calculating the standard deviation that uses the display buffer and will be displayed in the single channel view. This way, we can ensure both the mean and std values are accurately calculated and the offset correction functions properly.

Could you implement these changes and update the PR? If you’re unable to do so, please let us know and we’ll handle it. Thanks!

florian6973:
Hi!

Thank you very much for your detailed review! I did not check indeed "subtract offset"... I updated the PR accordingly.

Please let me know if there are any other fixes to add :)

anjaldoshi (approved):
Thanks for making the requested changes. Looks great! We'll let you know when a new release is out with these changes.

Issue #556, opened 2023-02-23
How to add chunked samples to buffer via DataBuffer::addToBuffer?


StoneLin0708:
I'm trying to add more than one sample to databuffer at once, but I end up getting it to work by making some changes to the implementation of `DataBuffer::addToBuffer`.
Specifically, change [Source/Processors/DataThreads/DataBuffer.cpp#L91](https://github.com/open-ephys/plugin-GUI/blob/b28c1bde94fbabcc92ab7c2222496643ee6abdf8/Source/Processors/DataThreads/DataBuffer.cpp#L91) from `data + (idx * numChans) + chan,` into `data + (chan * numItems) + idx`

```
    for (int i = 0; bs[i] != 0; ++i)
    {                                // for each of the dest blocks we can write to...
        blkIdx = 0;
        for (int j = 0; j < bs[i]; j+= chunkSize) 
        {                     // for each chunk...
            cSize = chunkSize <= bs[i] - j ? chunkSize : bs[i] - j;     // figure our how much you can write
            for (int chan = 0; chan < numChans; ++chan)         // write that much, per channel
            {
                buffer.copyFrom (chan,                           // (int destChannel)
                                 si[i] + j,                      // (int destStartSample)
L91 =>                           data + (chan * numItems) + idx, // (const float* source)
                                 cSize);                         // (int num samples)
            }
```

My changes are intended to copy "cSize" samples into channel "chan" from a continuous memory.
To achieve this, the memory layout should be organized as a two-dimensional matrix where each row represents the next "numItems" samples of a channel. Here is an example table to illustrate this:
|data|0|1|2|...|idx|...|numItems|
|-|-|-|-|-|-|-|-|
|0|0|1|2|...|idx|...|numItems-1|
|1|0+numItems|1+numItems|2+numItems|...|idx+numItems|...|numItems-1+numItems|
|2||||||||
|...||||||||
|chan|||||idx+chan*numItems|||
|...||||||||
|numChans||||||||

I believe that my modified code will correctly copy the desired samples into the buffer using this memory layout.
However, I would appreciate it if someone could review my changes and verify that they are correct.
Thank you in advance for your help!

jsiegle:
[...]

PR #639, opened 2024-12-06
Issue: submitting interleaved or sequential data (chunkSize != 1) to DataThread/DataBuffer

kevkdev:
Resolution to thread: https://github.com/open-ephys/plugin-GUI/issues/556

The _DataBuffer::addToBuffer_ method, as written, only copies data correctly if _chunkSize_ = 1. 

Fixes included:

1) Allowing _chunkSize_ of any value that is a multiple of _numItems_ (it wouldn't make sense, otherwise).
2) Handling the case where a _chunkSize_ copy is split between end-of-buffer and start-of-buffer for the circular buffer.
3) Optimizations of when contiguous data is copied; in particular with the vectors _timestamps_ _sequenceNumbers_, and _eventCodes_.
4) Added clarity to the functionality of _addToBuffer_, with examples, to the header comment for the method.

Testing was performed using a sin wave as input across multiple channels in a _DataThread::updateBuffer_ method:

```
  static int cntr{0};
  static int index{0};
  static int groups[] = {1, 2, 5, 10};

  std::vector<float> raw_samples;
  raw_samples.reserve(NUMBER_CHANNELS * MAX_SAMPLES_PER_CHANNEL);

  if (cntr++ == mSampleRate * 3) {
    if (++index == 4) {
      index = 0;
    }
    cntr = 0;
  }

  float* buf = raw_samples.data();
  auto grouping{groups[index]};

  for (auto i = 0; i < MAX_SAMPLES_PER_CHANNEL; i += grouping) {

    for (int chan = 0; chan < NUMBER_CHANNELS; ++chan) {
      for (auto j = 0; j < grouping; j++) {
        *buf++ = ::sinf(mSample[chan]) * 40;
        mSample[chan] += static_cast<float>(M_PI) * 2.0f / (mSampleRate * chan+1);
      }

    }
  }

  mDataBuffer->addToBuffer(raw_samples.data(), sample_numbers, timestamps,
    event_codes, MAX_SAMPLES_PER_CHANNEL, grouping);
```

MAX_SAMPLES_PER_CHANNEL was set to 10, the buffer size to 1024. This was done to create a scenario where a _chunkSize_ copy could be split across the circular buffer.

PS - Although I've been developing code for over 30 years, this is my first contribution to open source. If in any of this process, I've made a faux pas, provide grace and allow it to be a teaching moment. Thanks for considering this update to your project.

### DIFF
--- a/Source/Processors/DataThreads/DataBuffer.cpp
+++ b/Source/Processors/DataThreads/DataBuffer.cpp
@@ -68,45 +68,104 @@ int DataBuffer::addToBuffer (float* data,
                              int numItems,
                              int chunkSize)
 {
-    int startIndex1, blockSize1, startIndex2, blockSize2;
+    int bs[] = { -1, -1, 0 };    // zero-termintated as check in outer loop
+    int si[] = { -1, -1};
+    int cSize = chunkSize;
+    int offset = 0;
+    
+    abstractFifo.prepareToWrite (numItems, si[0], bs[0], si[1], bs[1]);
 
-    abstractFifo.prepareToWrite (numItems, startIndex1, blockSize1, startIndex2, blockSize2);
+    // The contiguous case is separated from the interleaved version to reduce the
+    // overhead of the for loops and small copies.
+    if (numItems == chunkSize)
+    {
+        for (int chan = 0; chan < numChans; ++chan)
+        {
+            // copy first half
+            buffer.copyFrom (chan,                       // destination channel
+                             si[0],                // offset into buffer
+                             data + chunkSize * chan,    // source data location
+                             bs[0]);                // number of samples
 
-    int bs[3] = { blockSize1, blockSize2, 0 };
-    int si[2] = { startIndex1, startIndex2 };
-    int cSize = 0;
-    int idx = 0;
-    int blkIdx;
-        
-    for (int i = 0; bs[i] != 0; ++i)
-    {                                // for each of the dest blocks we can write to...
-        blkIdx = 0;
-        for (int j = 0; j < bs[i]; j+= chunkSize) 
-        {                     // for each chunk...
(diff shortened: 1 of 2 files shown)


anjaldoshi:
Hi @kevkdev,

Thanks for submitting this PR! Although your fixes sound helpful, we’ve already taken care of this on our end. The change we made is in this commit: 7dd337d. This change is currently only available in v1.0.0 of the GUI, which is in preview. We plan to release the stable version soon. Since this change violates the Plugin API, we won’t be able to backport it.

Our modifications to the `DataBuffer::addToBuffer` method now employ a channel-major layout, which enhances data storage and processing. The key changes include:

- Removal of the chunkSize parameter, simplifying data handling.
- Updates to the addToBuffer function facilitate efficient data copying in contiguous blocks. By utilizing `AbstractFifo`, the implementation of circular buffering for multi-channel data is simplified, encompassing timestamps, sample numbers, and event codes.
- Reorganization of buffer internals to optimize memory and computation performance.

The refactor enhances clarity and operational speed by focusing on block-based copying rather than sample-based copying.

kevkdev:
I need  to pull the updated branch, but I believe the code has one of the issues I pointed out above - that given a chunk size that is not divisible by the size of the buffer, the copy method drops part of the data. If I get a chance in the next few days, I will do so.

Also, sadly, the data I receive from a remote entity has the channels interleaved (1 2 3 1 2 3) and not blocked (1 1 1 2 2 2 3 3 3). With that change, there's an additional copy from the interleaved to blocked data. 

Having a solution like the above wouldn't limit the incoming data sets, but allow for either of the two above, or some variants (e.g. 1 1 2 2 3 3 1 1 2 2 3 3), while maintaining the original API for this method, with optimizing towards the fewest number of copies. 

Just something to consider.

PR #656, opened 2025-04-27
[WIP] Record Node directory - offer permanent link to control panel

d1manson:
As far as i understand the control panel default path is copied to the record node when it's instantiated. But if the control panel changes after that the record node does not change.  Assuming i've understood that correctly, I do find that confusing.

Here I propose the path parameter on the record node has `isRequired` set to `false`, and the value set to `None` by default. The`None` value is rendered like this:
Screenshot 2025-04-27 at 23 52 47

When you click that, you get the existing folder picker which lets you pick a custom directory. If you have selected a custom dir, it looks no different to prior to the PR:
Screenshot 2025-04-27 at 23 54 35

However, when you mouse over the path you now get the option to clear it:
Screenshot 2025-04-27 at 23 49 30

(if you click the 'x' it goes back to the `None` state, otherwise if you click the main part of the path it opens the file picker as before).

I think I've implemented the UI aspect of this correctly, but the actual dataDir stuff I don't think I've quite got right as it needs to either listen for changes in the control panel (i haven't checked how to do this), or when the recording starts it should check at that point. I make look myself shortly, but welcome to suggestions on the right way to do that.

I appriciate this is not the most exciting feature, and I have dived straight in with a PR rather than discussing, but I won't be offended if it doesn't quite make sense in terms of UX and/or implementation.

### DIFF
--- a/Source/Processors/Parameter/Parameter.cpp
+++ b/Source/Processors/Parameter/Parameter.cpp
@@ -1230,6 +1230,9 @@ void PathParameter::setNextValue (var newValue_, bool undoable)
         else if (isDirectory && File (newValue_.toString()).exists())
         {
             newValue = newValue_;
+        } else if (!isRequired && newValue_.toString() == "None")
+        {
+            newValue = newValue_;
         }
 
         if (! undoable)
--- a/Source/Processors/Parameter/Parameter.h
+++ b/Source/Processors/Parameter/Parameter.h
@@ -962,11 +962,15 @@ public:
     /** Returns true if the path should be a directory */
     bool getIsDirectory() { return isDirectory; }
 
+    /** Returns true if the path is required */
+    bool getIsRequired() { return isRequired; }
+
     /** Returns a list of valid file extensions if applicable */
     StringArray getValidFilePatterns() { return filePatternsAllowed; }
 
     /** Returns true if the current path is valid for this parameter */
     bool isValid() override;
+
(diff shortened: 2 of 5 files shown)


jsiegle:
Thanks for implementing this! I agree that the default behavior can be a bit confusing, so it's good to have an option to change it. Are there any changes you still want to make before we merge this? We can help clean up things on the backend.

d1manson:
thanks for the quick reply @jsiegle .

it's totally up to you if you want to merge at this point even though the "backend" isn't fully implemented. I have had a bit of a look at the backend side of things just now, but it's not immediately clear to me the right approach.

in terms of code quality, I definitely wouldn't say this change is a net positive...in particular it seems a bit odd to use the special string `"None"` in the way I have done, and also to have the special string `"default"` kicking around.  Maybe one could use a juce `var::undefined` or `void`, rather than a string to represent `None`, and dispense with the `"default"` concept entirely..it also feels like the `isRequired` flag could be moved into the base `Paramater` class, as there's nothing path-specific about it. ..but those feel like more signficant changes that I didn't want to go near without having more experience in the codebase (tbh I only have limited experience with C++ in general)!

feel free to do as you wish with this! thanks again

jsiegle:
No problem! We appreciate your effort to make the Record Node work in a more intuitive way! I think others will be grateful for this as well.

PR #660, opened 2025-06-14
C++20: remove <T> from StreamSettings ctor/dtor

StoneLin0708:
C++20 forbids using a simple-template-id in constructor/destructor declarators (https://eel.is/c++draft/diff.cpp17.class#2).
Dropping the redundant `<T>` lets Source/Processors/Settings/DataStream.h compile with -std=c++20+. No functional change.


anjaldoshi:
Looks good! We've verified that merging this won't affect the API, so existing plugins will continue to function properly.

PR #669, opened 2025-08-22
Show the channel number in the tooltip, even if it is not shown on the left of the LFP viewer

MarinManuel:
I noticed that, when the channel height is small, channel numbers are replaced by `--` on the left of the LFP viewer. However, the text of the tooltip is also replaced by `--` which makes it impossible to know what channel one is looking at.

image

I made a simple change so that the channel number is always shown in the tooltip.
image


anjaldoshi (approved):
The tooltip should always display the channel number/name, regardless of its visibility on the left side of the channel plots. Thanks for catching and fixing this; it looks great!

I'll merge this into the `development` branch now, and the fix will be included in the next release of the GUI.

PR #682, opened 2025-12-22
Potential Vulnerability in Cloned Code

tabudz:
### Summary
Our tool detected a potential vulnerability in Resources/DeveloperTools/BinaryBuilder/JuceLibraryCode/modules/juce_core/zip/juce_ZipFile.cpp which was cloned from juce-framework/JUCE but did not receive the security patch applied. The original issue was reported and fixed under https://nvd.nist.gov/vuln/detail/cve-2021-23521.

### Proposed Fix
Apply the same patch as the one in juce-framework/JUCE to eliminate the vulnerability.

### Reference
https://nvd.nist.gov/vuln/detail/cve-2021-23521
https://github.com/juce-framework/JUCE/commit/2e874e80cba0152201aff6a4d0dc407997d10a7f

### DIFF
--- a/Resources/DeveloperTools/BinaryBuilder/JuceLibraryCode/modules/juce_core/files/juce_TemporaryFile.cpp
+++ b/Resources/DeveloperTools/BinaryBuilder/JuceLibraryCode/modules/juce_core/files/juce_TemporaryFile.cpp
@@ -105,7 +105,7 @@ bool TemporaryFile::deleteTemporaryFile() const
     // Have a few attempts at deleting the file before giving up..
     for (int i = 5; --i >= 0;)
     {
-        if (temporaryFile.deleteFile())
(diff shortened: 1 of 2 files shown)


anjaldoshi:
Hi @tabudz,

Thanks for taking the time to open this PR and flag the issue.  

The file you linked is part of an old BinaryBuilder tool that was only used to generate BinaryData for GUI assets (icons, fonts, etc.), and that version of JUCE is never compiled into or linked by the main GUI application, so it isn’t part of the shipped binary.  We’re planning to remove this old BinaryBuilder code from the repository soon to avoid this kind of confusion in the future.

The JUCE code that is actually used by the main GUI lives [here](https://github.com/open-ephys/plugin-GUI/tree/main/JuceLibraryCode) and is currently on JUCE v8.0.7, which does not include this particular security vulnerability.

PR #704, opened 2026-06-22
LFP Viewer support for low sampling rates

### LINKED ISSUE #703: LFP Viewer issues at low sampling rates
**Describe the bug**
LFP Viewer fails to plot signals with low sampling rates. The fix proposed in #532 is not sufficient below 100Hz. At lower sampling rates, the plot becomes increasingly irregular, the timebase drifts. Display completely fails at 1Hz.

**To Reproduce**
To make the error reproducible without any hardware device, I used the LSL Inlet module, and provided a low sampling rate data stream from Matlab (2024b).

The following Matlab script produces a 0.1Hz sine wave sampled at 1Hz as a single channel LSL stream, via [liblsl-Matlab](https://github.com/labstreaminglayer/liblsl-Matlab).

```matlab
fs = 1;  % sampling rate (Hz)

lib = lsl_loadlib();

info = lsl_streaminfo(lib, 'Test_stream', 'misc', 1, fs, 'cf_float32', '');
outlet = lsl_outlet(info);

disp('Streaming Test_stream. Press Ctrl+C to stop.');

sample_idx = 0;
while true
  t = sample_idx / fs;  % seconds elapsed
  sample = 3500 * sin(2 * pi * 0.1 * t);  % 0.1 Hz, amplitude 3500.0 uV
  outlet.push_sample(sample);
  sample_idx = sample_idx + 1;
  pause(1 / fs);
end
```

**Screenshots**
The attached screenshots show the visualisation of the same 0.1Hz sine wave at various sampling rates (1000Hz, 100Hz, 50Hz, 10Hz, 1Hz).

Image
Image
Image
Image
Image

**Versions:**
 - OS: macOS 14.7.5, Windows 11 (25H2)
 - GUI version: 1.0.2
 - Plugin version: LSL Inlet 1.0.0-API10

**Additional context**
I am interested in collecting slow physiological signals along with neural activity. Auxiliary sensor data is received in Open Ephys via the LSL Inlet plugin. The issue is unrelated to the sensor device, as it can be reproduced on synthetic data streams as above.

roberttoth02:
Resolves #703

1. Ensures very low sampling rate streams are actually plotted along the window by introducing minimum buffer length
2. Fixes time axis drift at low sampling rates by revising the subSampleOffset accumulation mechanism
3. Adds linear interpolation between datapoints where points / pixels < 1.0
4. Adds new unit test group targeting the above fixes

### DIFF
--- a/Plugins/LfpViewer/DisplayBuffer.cpp
+++ b/Plugins/LfpViewer/DisplayBuffer.cpp
@@ -27,6 +27,7 @@ namespace LfpViewer
 {
 
 #define BUFFER_LENGTH_S 1.0f
+#define MIN_BUFFER_SAMPLES 1000
 
 DisplayBuffer::DisplayBuffer (int id_, String name_, float sampleRate_) : id (id_), name (name_), sampleRate (sampleRate_), isNeeded (true)
 {
@@ -104,7 +105,7 @@ void DisplayBuffer::addChannel (
 void DisplayBuffer::update()
 {
     if (numChannels != previousSize)
-        setSize (numChannels + 1, int (sampleRate * BUFFER_LENGTH_S));
+        setSize (numChannels + 1, std::max (int (sampleRate * BUFFER_LENGTH_S), MIN_BUFFER_SAMPLES));
 
     clear();
 
--- a/Plugins/LfpViewer/LfpDisplayCanvas.cpp
+++ b/Plugins/LfpViewer/LfpDisplayCanvas.cpp
@@ -1357,6 +1357,11 @@ void LfpDisplaySplitter::updateScreenBuffer()
 
             int sampleNumber = 0;
 
+            // For ratio < 1: save the display-buffer start index so we can
+            // compute each pixel's sample position with a direct multiply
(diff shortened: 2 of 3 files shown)


anjaldoshi (approved):
Hi @roberttoth02,

Thanks a lot for digging into this and fixing the low-sample-rate issues in the LFP Viewer.

This PR looks great overall, and thanks for targeting the `development` branch. That helps a lot since we already have a few other LFP Viewer changes there. The timing is great too, since we’re hoping to ship `v1.1.0` this week and we’d love to get this in.

I went through the code and the approach makes a lot of sense. We just haven’t had much opportunity to exercise the viewer with very low sample-rate data before, especially below 100 Hz.

I also tested this locally with the LSL Inlet plugin using `pylsl` to stream sine-wave data at low sampling rates. Just like you described in the issue, the current `development` behavior drifts quite a bit once the samples-per-pixel ratio drops below 1, and at 1 Hz it basically fails. With your changes, the drift looks fixed and the traces render much more naturally with the interpolation.

The new tests are a really helpful addition too.

Overall, this is a solid fix, and I’m glad we can get it into `v1.1.0`.

